PR #7, opened 2026-01-16
Implement EIP-8025 guest program design

jsign:
This PR implements the current [EIP-8025 draft](https://github.com/ethereum/consensus-specs/pull/4828). In this repo, today we support Ethrex and Reth so both work -- but this code can help other ELs to understand how EIP-8025 expects guests to work.

It is a big-ish PR, so I’ll summarize the key points here to keep the important parts concise and make the review feel natural (no quirks/hacks). Still, I will soon create PR comments to revisit the points below and help as much as possible with the review. (I think it is quite hard to do this in multiple PRs). I'll also create some issues to track many comments.

TL;DR:
- Previously, the guest program's private input included the EL block (+exec witness). Now it receives `NewPayloadRequest` (+exec witness) instead. We support multiple `NewPayloadRequest` versions since this change per fork (I use an enum).
- The public input switches from `(parent_hash, block_hash, valid/invalid)` to `(tree_hash_root(NewPayloadRequest), valid/invalid)`
- Important missing protocol validations are now present:
    - Before, we received the EL block. This didn’t allow the guest program to perform many important checks outside the STF logic without asking CLs to do them (which requires RLP and similar)
    - Now we do the full checking of the EngineAPI layer e.g. `block_hash` check between the `NewPayloadRequest` and derived block, CL `versioned_blob_hashes` against the present ones in the EL transactions, etc.
- Guest program logic is wrapped to catch panics and have the expected `(tree_hash_root(NewPaylaodRequest), false)` — this is a technical detail that is preferred now until there is more clarity reg panic behavior in zkVMs, since we have the `true/false` field in public inputs.

## Performance impact

Comparing the previous version to the current one is like comparing apples to oranges — in the previous version, we didn’t have many checks we knew were needed. Still, I ran the previous and new versions of Reth/Zisk over 50 mainnet blocks using an 8x5090 prover.

The average speedup is 0.96x — much better than I expected. For the SSZ hashing, the SHA2 zkVM precompile prob helped a lot, and I paid attention to our `NewPayloadRequest` transformation to the underlying EL block has as few allocations/copy as possible — this is usually possible since there is a quite direct map between our standardized `NewPayloadRequest` and the underlying EL types (just had to be lucky enough to transform types with moves instead of copy).

## Mid-level important comments

Other important comments.

### Regarding how to ingest `NewPayloadRequest` into the guest program
- `NewPayloadRequest` is not a data structure related to EL RPCs. Ideally, I wanted a way to still be able to generate proofs only relying on EL RPC (i.e. `eth_getBlock` + `debug_executionWitness`). We could rely on CL **and** EL RPC, but that would add an extra RPC dependency and introduce more edge cases with unsynced nodes.
- To still allow, only requiring EL RPCs for proof creation (mostly for optional proofs, not block building for obvious reasons). I created some helper code that derives the `NewExecutionPayload` from an EL block. It is not enough to “map fields,” since, for example, the `request_bytes` aren’t present in the EL block, but only in the root. To solve this, I execute the block on the host, pull the' request_bytes' result, and use it to fill `NewExecutionPayload`. For optional proofs, this adds an extra execution in the host, but we could eventually include `request_bytes` in a newly designed `debug_executionWitnessV2` to avoid the extra execution (should be easy).
- The last point means that zkboost and workload repos are unaffected by this change in their needs to generate private inputs, which I wanted to preserve to simplify infrastructure.

### Reg CL/EngineAPI types for `NewPayloadRequest` and other downstream types
- I decided not to leverage any existing crate for CL types like e.g. `NewPayloadRequest`, `ExecutionPayloadV{1,2,3}`, `Withdawals`, etc.
- The reason is that existing ones like Lighthouse are in pretty big `types` crate, which contains other CL types that need complicated dependencies for zkVMs (e.g. blst, kzg) without an easy way to feature-gate them.
- I decided to do our own minimal version of exactly what we need, with as few dependencies as possible.
- Additionally, we want to eventually ELs (e.g. Ethrex, Reth in repo today) to upstream the code I wrote reg `NewPayloadRequest->EL block` of the guest into their codebases (since we use Rust here at least). I doubt they will accept using Lighthouse crates if that might pull many other deps for them. I wouldn't complain at all if the Lighthouse team is okay with extracting what we need into a separate, minimal crate so we don't have to maintain our own :)
- Added support for both serde and rkyv in these minimal-CL types I did, since they are used by Reth and Ethrex, respectively. This was to allow them to keep their preferred serialization of private inputs. Needed a few custom serde/rkyv wrappers to make some `ssz_types` work.
- Bottom line: the received `NewPayloadRequest` is the same for Ethrex and Reth, compared to our previous reality, where each received their own “Block type”. This is nice since we’re moving closer to them receiving a standardized input.

### SSZ
- For SSZ, I use `ethereum_ssz` from Lighthouse, since it is properly isolated from kzg and blst dependencies. Although every code is critical for the guest program, def I think we don't need new SSZ impl if possible. The only side effect is that it is not `no_std` which is bad for Airbender. I spent a while patching crates, but I already needed to patch 6 crates before I felt too uncomfortable having so many patches -- prob worth revisiting in a separate PR again to double-check if there's a simpler way.

### Reg public input verification
- Since now the public input is `tree_hash_root(NewPayloadRequest)` instead of our previous `block_hash`, there wasn’t a clear way to verify that the guest program execution is correct since this root isn’t part of EL RPC (`eth_getBlock` and `debug_executionWitness`). As in, there isn’t a clear way to generate the expected public input without relying on EL guest execution itself.
- To be completely sure things are running as expected, I had to write an independent CLI tool based on Lighthouse to calculate this root for any slot. This is to have an independent value to compare against in our existing integration tests mainnet blocks.
- We have 15 mainnet blocks in integration tests. I looked at which slots they correspond to, and used this tool in my Lighthouse CLI to generate the `hash_tree_root(NewPayaloadRequest)`. Using this as an independent source of truth, I compare EL guest execution public input against this map.
- This has helped me find many bugs during the implementation. I could bisect why the roots were unmatching by looking deeper into the internal roots and catching/fixing bugs.
- This tool is quite dirty — at some point, I’ll clean it up and put it somewhere just in case we change the blocks in the integration tests.

### Crate patching
- The only crate I had to patch was `ethereum_hashing`, which is an indirect dependency from `ethereum_ssz`  if you rely on `TreeHash` derive. `ethereum_hashing` is only used as a proxy to decide which SHA256 crate is used for SSZ hashing. It unconditionally pulled `ring` crate, which uses a C-lib -- not feature gated, unfortunately. It has support for `sha2` (which we want since most zkVM precompiles patch it) but only enabled by CPU-feature detection, and not a direct way through feature-gating. In any case, it is an extremely simple patch and will require zero maintenance (but we could/should upstream at some point).

### Unmodified Reth and Ethrex repos
- This was important IMO, since forking ELs is hairy. I was very lucky that Reth recently exposed the block execution output in the stateless validator function, which allowed me to capture the generated requests (for the host).
- Unfortunately-ish, for Reth and Ethrex I had to pull some code in our codebase for EngineAPI checks reg block and payload. Their corresponding crates had networking dependencies or C-libs. This is because they usually live in their Engine API crates, which makes sense.
- For Reth, I only had to copy paste one small function — but we need to create a PR to add some features to gate some dependencies to remove this in our code.
- For Ethrex, had to pull multiple functions, all quite simple. Still, we need to create a PR or ask them to extract these Payload/Block checks outside the Engine API crate to avoid network dependencies — should be very easy to do.
- TL;DR: In both cases, the code is minimal and simple, and I specifically isolated it so it is easy to delete once upstream has it in a nice shape to call directly. I preferred this to forking their codebases.

### What did all this mean for core guest program logic
The only change that the existing guest program logic has compared to before this PR is:
- Transform `NewPayloadRequest` from private input into their corresponding EL block type (for Ethrex and Reth are distinct types ofc). I did this with a method that tries to avoid as many allocations/copies as possible and leverage existing code in their codebases (since they do this in the EngineAPI). As part of that, they have to do the Payload/Block validation checks I mentioned in the TL;DR.
- Do the `NewPayloadRequest.tree_hash_root()` for the public input — very easy and transparent from the guest program perspective.
- All the rest is the same, since once you have the EL block, you do the same STF checking logic as before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1030,7 +1030,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -2703,7 +2703,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
@@ -3186,7 +3186,7 @@ dependencies = [
 
 [[package]]
 name = "integration-tests"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-primitives",
  "block-encoding-length",
@@ -6031,7 +6031,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
 name = "stateless-validator-common"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -6050,7 +6050,7 @@ dependencies = [
 
 [[package]]
 name = "stateless-validator-ethrex"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -6074,7 +6074,7 @@ dependencies = [
 
 [[package]]
 name = "stateless-validator-reth"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ resolver = "2"
 edition = "2024"
 license = "MIT OR Apache-2.0"
 rust-version = "1.88"
-version = "0.3.0"
+version = "0.4.0"
 
 [workspace.lints.rust]
 missing_debug_implementations = "warn"
--- a/bin/block-encoding-length/airbender/Cargo.lock
+++ b/bin/block-encoding-length/airbender/Cargo.lock
@@ -618,7 +618,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1305,7 +1305,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/block-encoding-length/openvm/Cargo.lock
+++ b/bin/block-encoding-length/openvm/Cargo.lock
@@ -618,7 +618,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1314,7 +1314,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/block-encoding-length/pico/Cargo.lock
+++ b/bin/block-encoding-length/pico/Cargo.lock
@@ -799,7 +799,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1902,7 +1902,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/block-encoding-length/risc0/Cargo.lock
+++ b/bin/block-encoding-length/risc0/Cargo.lock
@@ -812,7 +812,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1604,7 +1604,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/block-encoding-length/sp1/Cargo.lock
+++ b/bin/block-encoding-length/sp1/Cargo.lock
@@ -646,7 +646,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1348,7 +1348,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/block-encoding-length/zisk/Cargo.lock
+++ b/bin/block-encoding-length/zisk/Cargo.lock
@@ -628,7 +628,7 @@ dependencies = [
 
 [[package]]
 name = "block-encoding-length"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1318,7 +1318,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
--- a/bin/stateless-validator-ethrex/risc0/Cargo.lock
+++ b/bin/stateless-validator-ethrex/risc0/Cargo.lock
@@ -1923,7 +1923,7 @@ dependencies = [
 
 [[package]]
 name = "guest"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "ere-io",
  "ere-platform-trait",
@@ -3895,7 +3895,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
 name = "stateless-validator-common"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "anyhow",
  "ethereum_ssz",
@@ -3912,7 +3912,7 @@ dependencies = [
 
 [[package]]
 name = "stateless-validator-ethrex"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "anyhow",
  "bytes",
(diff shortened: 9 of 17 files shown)


crates/integration-tests/src/lib.rs
```diff
@@ -134,3 +151,22 @@ impl TestCase {
...
+    }
+
+    fn write_whole_output(_: &[u8]) {
+        panic!("Can't write output in NoopPlatform");
+    }
+
+    fn print(message: &str) {
+        println!("{}", message);
+    }
+}
```
jsign: This is just a helper ere Platform used to execute the guest program on the host.

crates/stateless-validator-common/src/new_payload_request.rs
```diff
@@ -0,0 +1,321 @@
...
+            NewPayloadRequest::Bellatrix(req) => req.tree_hash_root().0,
+            NewPayloadRequest::Capella(req) => req.tree_hash_root().0,
+            NewPayloadRequest::Deneb(req) => req.tree_hash_root().0,
+            NewPayloadRequest::ElectraFulu(req) => req.tree_hash_root().0,
+        }
+    }
+}
+
+/// Computes the requests hash for EL block construction.
+pub fn compute_requests_hash(requests: &ExecutionRequests) -> [u8; 32] {
```
jsign: This is the neutral impl of the inverse of the `decode_execution_requests(...)` comment I did before. Inside the guest, they receive `NewPaylaodRequest` and need to construct the `requests_hash` for their EL block type.

crates/stateless-validator-ethrex/src/execution_payload.rs
```diff
@@ -0,0 +1,192 @@
...
+    Address, Bloom, H256,
+    constants::DEFAULT_OMMERS_HASH,
+    types::{
+        Block, BlockBody, BlockHeader, Transaction, Withdrawal, compute_transactions_root,
+        compute_withdrawals_root,
+    },
+};
+use ethrex_rlp::error::RLPDecodeError;
+
+#[derive(Clone, Debug)]
```
jsign: For Ethrex, most of this file are some things I had to pull since they live in crates that have complicated dependencies like network ones (since they live in the EngineAPI crate). The same happen with methods that you see here.

I think the goal is to ask the Ethrex team to move these types into their codebase to nicier crates, with that we can nuke this module entirely and rely directly on them eventually.

For Reth there is a similar situation, but had to only pull one method -- I'll create a comment there.

crates/integration-tests/tests/stateless-validator-ethrex.rs
```diff
@@ -1,31 +1,22 @@
...
-                fixture.success,
-            );
-            TestCase::new::<StatelessValidatorEthrexGuest>(fixture.name, input, output)
-                .output_sha256()
-        });
+    let fixtures = get_fixtures();
+    let inputs = fixtures.into_iter().map(|fixture| {
+        let input =
+            StatelessValidatorEthrexInput::new(&fixture.stateless_input, fixture.success).unwrap();
+        let output = StatelessValidatorEthrexGuest::compute::<NoopPlatform>(input.clone());
```
han0110: Should we use the same `get_stateless_validator_output` here like what reth did?

crates/stateless-validator-ethrex/src/new_payload_request.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        NewPayloadRequest::ElectraFulu(e) => {
+            let parent_beacon_block_root = Some(H256::from(e.parent_beacon_block_root));
+            let requests_hash = Some(H256::from(compute_requests_hash(&e.execution_requests)));
+            let payload: ExecutionPayload = e.execution_payload.into();
+            validate_execution_payload_v3(&payload).context("V3 payload validation failed")?;
+            payload
+                .into_block(parent_beacon_block_root, requests_hash)
+                .context("into_block failed")
+        }
```
han0110: Why is `validate_block_payload_v3` not needed here? I think the `versioned_hashes` is not used for this case.

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -17,21 +17,38 @@ pub use {
...
+    pub new_payload_request: NewPayloadRequest,
+    /// database containing all the data necessary to execute
+    pub execution_witness: ExecutionWitness,
+    /// value used to calculate base fee
+    pub elasticity_multiplier: u64,
+    /// Configuration for L2 fees used for each block
+    pub fee_configs: Option<Vec<FeeConfig>>,
+    #[cfg(feature = "l2")]
+    /// KZG commitment to the blob data
+    #[serde_as(as = "[_; 48]")]
```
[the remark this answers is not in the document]
jsign: Ah, these are just original code from Ethrex [here](https://github.com/lambdaclass/ethrex/blob/ff00c885b79662f6d64af38b1632e14bc7cd7614/crates/l2/prover/src/guest_program/src/input.rs#L25-L29). These fields are just relevant for their `l2` feature which we don't use -- but I think now that we defined our own type I can remove them directly.

crates/integration-tests/tests/stateless-validator-ethrex.rs
jsign: Ah yes, this was the idea. Changed it here https://github.com/eth-act/ere-guests/pull/7/commits/00a7957608b1d92849f42c567f96d9610ad15f57

crates/stateless-validator-common/src/new_payload_request.rs
```diff
@@ -0,0 +1,321 @@
...
+    feature = "rkyv",
+    derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
+)]
+pub struct NewPayloadRequestFulu {
+    pub execution_payload: ExecutionPayloadV3,
+    #[cfg_attr(feature = "rkyv", rkyv(with = crate::rkyv_wrappers::AsVariableList))]
+    pub versioned_hashes: VariableList<Hash32, MaxBlobCommitmentsPerBlock>,
+    pub parent_beacon_block_root: Hash32,
+    pub execution_requests: ExecutionRequests,
+}
```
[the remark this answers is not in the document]
jsign: Ah, yes. I had my own back and forth reg this unification for Electra and Fulu. After leaning on the unification I think this was a left over -- removed here https://github.com/eth-act/ere-guests/pull/7/commits/02b04d661a601064bda4bc31d25a79d2a495be02

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -76,44 +81,96 @@ pub struct StatelessValidatorEthrexGuest;
...
+        };
+        let input = ProgramInput {
+            blocks: vec![block],
+            execution_witness: input.execution_witness,
+            elasticity_multiplier: input.elasticity_multiplier,
+            fee_configs: input.fee_configs,
+            #[cfg(feature = "l2")]
+            blob_commitment: input.blob_commitment,
+            #[cfg(feature = "l2")]
+            blob_proof: input.blob_proof,
```
han0110: could be cleanup

PR #16, opened 2026-02-02
Update Reth and Ethrex

jsign:
Updates both Reth and Ethrex to their latest release:
- Reth v1.10.2
- Ethrex to v9.0.0.


jsign:
Github actions is having a bad moment now -- I'll retry CI later.

PR #26, opened 2026-03-20
Switch to libssz

### LINKED ISSUE #8: Make `stateless-validator-common` no_std
As mentioned in the PR description, `ethereum_ssz` has some chain of crates dependencies which aren't `no_std`. Went into that rabbit hole a bit, and got pretty far but required at least these patches:
```
ethereum_ssz = { git = "https://github.com/han0110/ethereum_ssz", branch = "feature/no-std" }
ethereum_ssz_derive = { git = "https://github.com/han0110/ethereum_ssz", branch = "feature/no-std" }
ethereum_serde_utils = { git = "https://github.com/han0110/ethereum_serde_utils", branch = "feature/no-std" }
tree_hash = { git = "https://github.com/jsign/tree_hash", rev = "2263d501c0765837ef099796f153ba323939e614" }
tree_hash_derive = { git = "https://github.com/jsign/tree_hash", rev = "2263d501c0765837ef099796f153ba323939e614" }
ssz_types = { git = "https://github.com/jsign/ssz_types", rev = "679b7ba3e7402df8f8441af2242a6117c91fc84a" }
````
The first three are ones that we already did before for similar reasons, and bottom three are ones I had to do because of new needs. I got pretty close -- but there is still some extra things to do potentially. I prefered to stop and not include it in this PR since it was already quite big; but I think we can revisit this to see if we need to continue patching or there is other way.

_Originally posted by @jsign in https://github.com/eth-act/ere-guests/pull/7#discussion_r2698358245_

### PR DESCRIPTION
**Pull request by jsign**
This PR switches from Ligthouse SSZ library to Lambdaclass new [SSZ library](https://github.com/lambdaclass/libssz). 

This has many benefits:
1. `libssz` has native `no_std` support, so it allows us to now run Airbender.
2. From the measurements I've done with Zisk cost calculator, the hash tree root for the output commitment is 2x faster. Technically, this should result in an overall proving speedup of ~5% (previous hash tree root influence in total proving time was ~10%).
3. Much less indirect dependencies, and also remove one `ethereum_hashing` patch I had to do, so less complexity/forks.

Fixes https://github.com/eth-act/ere-guests/issues/8

### DIFF
--- a/crates/stateless-validator-common/src/rkyv_wrappers.rs
+++ b/crates/stateless-validator-common/src/rkyv_wrappers.rs
@@ -1,6 +1,6 @@
-//! rkyv wrappers for ssz_types (VariableList, FixedVector).
+//! rkyv wrappers for `libssz_types`.
 //!
-//! ssz_types doesn't support rkyv natively, so we provide wrappers that
+//! `libssz_types` doesn't support rkyv natively, so we provide wrappers that
 //! serialize them as `Vec<T>` and reconstruct on deserialization.
 
 use alloc::{format, string::String, vec::Vec};
@@ -13,10 +13,9 @@ use rkyv::{
     vec::{ArchivedVec, VecResolver},
     with::{ArchiveWith, DeserializeWith, SerializeWith},
 };
-use ssz_types::{FixedVector, VariableList};
-use typenum::Unsigned;
+use libssz_types::SszList;
 
-/// Simple error wrapper for ssz_types errors which don't implement std::error::Error.
+/// Simple error wrapper for `libssz_types` errors which don't implement `std::error::Error`.
 #[derive(Debug)]
 struct SszError(String);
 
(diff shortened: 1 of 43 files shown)

### CODE REVIEW

File: crates/stateless-validator-common/src/serde_wrappers.rs [hunk 1]
```diff
@@ -25,3 +29,68 @@ pub mod bytes_array {
         serde_with::Bytes::deserialize_as(deserializer)
     }
 }
+
+/// Serialize an `SszList<T, N>` as `Vec<T>`.
+pub mod ssz_list {
+    use serde::{Deserialize, Serialize, de::Error};
+
+    use super::*;
+
```
kevaundray, line 38: was this supported in the original ssz library, but not libssz?

kevaundray: Ah there was a VariableList -- I wonder if we can upstream this or whether it is out of scope

jsign: Yeah not sure if is in scope for them -- maybe want to avoid bloating. 😅

crates/stateless-validator-common/src/rkyv_wrappers.rs
```diff
@@ -28,175 +27,97 @@ impl fmt::Display for SszError {
...
 impl core::error::Error for SszError {}
 
-/// Wrapper to serialize `VariableList<T, N>` as `Vec<T>`.
+/// Wrapper to serialize `SszList<T, N>` as `Vec<T>`.
 ///
 /// On serialization, the inner slice is serialized as a Vec.
-/// On deserialization, the Vec is converted back to VariableList.
+/// On deserialization, the Vec is converted back to `SszList`.
 #[derive(Debug)]
 pub struct AsVariableList;
```
han0110:
```suggestion
pub struct AsSszList;
```

crates/stateless-validator-common/src/rkyv_wrappers.rs
```diff
@@ -28,175 +27,97 @@ impl fmt::Display for SszError {
...
+/// Wrapper for nested `SszList` types like `SszList<SszList<u8, M>, N>`.
 ///
-/// This is used for fields like `Transactions = VariableList<Transaction, _>`
-/// where `Transaction = VariableList<u8, _>`.
+/// This is used for fields like `Transactions = SszList<Transaction, _>`
+/// where `Transaction = SszList<u8, _>`.
 ///
 /// Serializes as `Vec<Vec<T>>` and reconstructs on deserialization.
 #[derive(Debug)]
 pub struct AsNestedVariableList;
```
han0110:
```suggestion
pub struct AsNestedSszList;
```

han0110 (approved):
LGTM! Just a nit on naming consistency, it's good to keep as is.

PR #27, opened 2026-03-20
Update `ere` to `v0.5.0`

han0110:
Update `ere` to `v0.5.0`, which updates zkVMs:
- OpenVM from `v1.4.2` to `v1.4.3`
- SP1 from `v5.2.4` to `v6.0.1`
- ZisK from `v0.15.0` to `v0.16.0`

Because the patching crates of ZisK seems not available for `v0.16.0` and they are moving to use crypto backend for both Reth and Ethrex, so this PR also update Ethrex to an unreleased revision that has crypto backend supports. Also the `libssz` has `sha2` that's now not patched for ZisK, I'd expect the cost ratio of it will increase a bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,10 +91,10 @@ libssz_merkle = { package = "libssz-merkle", git = "https://github.com/lambdacla
 libssz_types = { package = "libssz-types", git = "https://github.com/lambdaclass/libssz", rev = "31786524708d853f5ac8da5bdfd8090a9b99103d", default-features = false, features = ["alloc"] }
 
 # ere
-ere-dockerized = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
-ere-io = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
-ere-platform-trait = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
-ere-zkvm-interface = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
+ere-dockerized = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
+ere-io = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
+ere-platform-trait = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
+ere-zkvm-interface = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
 
 # local
 guest = { path = "crates/guest", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ rust_2018_idioms = { level = "deny", priority = -1 }
 unnameable_types = "warn"
 unreachable_pub = "warn"
 unused_must_use = "deny"
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(zisk_hints)', 'cfg(zisk_hints_debug)', 'cfg(target_vendor, values("zisk"))'] }
 
 [workspace.lints.rustdoc]
 all = "warn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,11 +78,13 @@ stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "ed189a519
 tries = { git = "https://github.com/paradigmxyz/stateless", rev = "ed189a51931e8589102f3139d48fdbb3bbe4c1f7", default-features = false }
 
 # ethrex
-ethrex-common = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false, package = "guest_program" }
-ethrex-rlp = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-rpc = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-vm = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
+# FIXME: Update to upstream when https://github.com/lambdaclass/ethrex/pull/6392 is resolved.
+ethrex-common = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
+ethrex-crypto = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
+ethrex-guest-program = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
+ethrex-rlp = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
+ethrex-rpc = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
+ethrex-vm = { git = "https://github.com/han0110/ethrex.git", rev = "b9d3ef6924633ade5cc5e8f41b83f8ff7b3ec192", default-features = false }
 
 # ssz
 libssz = { git = "https://github.com/lambdaclass/libssz", rev = "31786524708d853f5ac8da5bdfd8090a9b99103d", default-features = false, features = ["alloc"] }
--- a/bin/stateless-validator-ethrex/sp1/Cargo.toml
+++ b/bin/stateless-validator-ethrex/sp1/Cargo.toml
@@ -23,10 +23,10 @@ crypto-bigint = { git = "https://github.com/sp1-patches/RustCrypto-bigint", tag 
 tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", tag = "patch-2.0.2-sp1-6.0.0" }
 p256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-p256-13.2-sp1-6.0.0" }
 secp256k1 = { git = "https://github.com/sp1-patches/rust-secp256k1", tag = "patch-0.30.0-sp1-6.0.0" }
-# ecdsa = { git = "https://github.com/sp1-patches/signatures", tag = "patch-16.9-sp1-6.0.0" }
 k256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-k256-13.4-sp1-6.0.0" }
 substrate-bn = { git = "https://github.com/sp1-patches/bn", tag = "patch-0.6.0-sp1-6.0.0-substrate-bn" }
 
+# NOTE: Commented out due to different `sp1-lib` version.
 # [patch."https://github.com/lambdaclass/bls12_381"]
 # bls12_381 = { git = "https://github.com/lambdaclass/bls12_381-patch/", branch = "expose-affine-constructors" }
 
--- a/bin/stateless-validator-ethrex/sp1/Cargo.toml
+++ b/bin/stateless-validator-ethrex/sp1/Cargo.toml
@@ -7,7 +7,7 @@ version = "0.1.0"
 
 [dependencies]
 # ere
-ere-platform-sp1 = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
+ere-platform-sp1 = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
 
 # local
 stateless-validator-ethrex = { path = "../../../crates/stateless-validator-ethrex", features = [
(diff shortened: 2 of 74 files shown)


bin/stateless-validator-reth/sp1/Cargo.toml
```diff
@@ -13,20 +13,22 @@ alloy-primitives = { version = "1.5.0", default-features = false, features = [
...
-ere-platform-sp1 = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
+ere-platform-sp1 = { git = "https://github.com/eth-act/ere", tag = "v0.5.0" }
 
 # local
 stateless-validator-reth = { path = "../../../crates/stateless-validator-reth" }
 
 [patch.crates-io]
-sha2-v0-10-9 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha2", tag = "patch-sha2-0.10.9-sp1-4.0.0" }
-sha3-v0-10-8 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha3", tag = "patch-sha3-0.10.8-sp1-4.0.0" }
-crypto-bigint = { git = "https://github.com/sp1-patches/RustCrypto-bigint", tag = "patch-0.5.5-sp1-4.0.0" }
```
han0110, line 25: I found that in [`rsp`](https://github.com/succinctlabs/rsp/blob/main/bin/client/Cargo.toml) they didn't patch `crypto-bigint`, do we remember why we have this?

bin/stateless-validator-ethrex/sp1/Cargo.toml
```diff
@@ -7,26 +7,28 @@ version = "0.1.0"
...
-ecdsa = { git = "https://github.com/sp1-patches/signatures", tag = "patch-16.9-sp1-4.1.0" }
-k256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-k256-13.4-sp1-5.0.0" }
-substrate-bn = { git = "https://github.com/sp1-patches/bn", tag = "patch-0.6.0-sp1-5.0.0" }
+sha2-v0-10-9 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha2", tag = "patch-sha2-0.10.9-sp1-6.0.0" }
+sha3-v0-10-8 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha3", tag = "patch-sha3-0.10.8-sp1-6.0.0" }
+crypto-bigint = { git = "https://github.com/sp1-patches/RustCrypto-bigint", tag = "patch-0.5.5-sp1-6.0.0" }
+tiny-keccak = { git = "https://github.com/sp1-patches/tiny-keccak", tag = "patch-2.0.2-sp1-6.0.0" }
+p256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-p256-13.2-sp1-6.0.0" }
+secp256k1 = { git = "https://github.com/sp1-patches/rust-secp256k1", tag = "patch-0.30.0-sp1-6.0.0" }
+# ecdsa = { git = "https://github.com/sp1-patches/signatures", tag = "patch-16.9-sp1-6.0.0" }
```
han0110: Not sure why but this is missing, and `sp1-patches/k256@patch-k256-13.4-sp1-6.0.0` is already using [`sp1-patches/ecdsa@sp1-skip-verify-on-recovery`](https://github.com/sp1-patches/signatures/tree/sp1-skip-verify-on-recovery) ([ref](https://github.com/sp1-patches/elliptic-curves/blob/patch-k256-13.4-sp1-6.0.0/k256/Cargo.toml#L35)), so I comment out the patch here.

bin/stateless-validator-ethrex/zisk/Cargo.toml
```diff
@@ -7,23 +7,17 @@ version = "0.1.0"
 
 [dependencies]
 # ere
-ere-platform-zisk = { git = "https://github.com/eth-act/ere", tag = "v0.3.0" }
+# NOTE: Using `default-features = false, features = ["inputcpy"]` mainly to disable the other
+#       feature enabled by default `user-hints`, which this binary doesn't use but introduces lots
+#       of dependnecy.
```
jsign:
```suggestion
#       of dependencies.
```

Cargo.toml
```diff
@@ -78,11 +79,13 @@ stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "ed189a519
 tries = { git = "https://github.com/paradigmxyz/stateless", rev = "ed189a51931e8589102f3139d48fdbb3bbe4c1f7", default-features = false }
 
 # ethrex
-ethrex-common = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false, package = "guest_program" }
-ethrex-rlp = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-rpc = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
-ethrex-vm = { git = "https://github.com/lambdaclass/ethrex.git", tag = "v9.0.0", default-features = false }
+# FIXME: Update to upstream when https://github.com/lambdaclass/ethrex/pull/6392 is resolved.
```
jsign, line 82: Let's wait for today and if isn't merged I'll ping Ethrex guys

crates/stateless-validator-ethrex/src/guest/zisk.rs
```diff
@@ -0,0 +1,22 @@
+// Copied and modified and modified from https://github.com/0xPolygonHermez/zisk-eth-client/blob/develop-0.8.0/crates/guest-ethrex/src/crypto/mod.rs.
```
jsign, line 1: Worth creating an issue to track this? We can ping hector there to help do required changes to use directly

crates/integration-tests/src/lib.rs
```diff
@@ -111,6 +111,10 @@ pub fn test_execution(
             expected_public_values.resize(32, 0);
         }
 
+        if matches!(zkvm_kind, zkVMKind::Zisk) && expected_public_values.len() < 256 {
```
jsign, line 114: I think these kind of padding we have to do today should probably go away after zkVMs comply the to the zkVM standards. Although I'm not entirely sure we are being very explicit about it as part of the standard.

Context (Han correct me if wrong): whatever the guest program wrote as public input isn't exactly the same thing we get from the SDK but padded. I think ideally we get exactly the same thing to not having to pad the expected output to what the SDK returns. cc @kevaundray @marcinbugaj

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -148,6 +142,18 @@ impl StatelessValidatorEthrexGuest {
     }
 }
 
+#[allow(unreachable_code)]
+fn crypto() -> Arc<dyn Crypto> {
```
jsign, line 146: Exciting to start seeing this. cc @kevaundray

bin/stateless-validator-ethrex/sp1/Cargo.toml
jsign: Man, seeing "skip verify" on a crate patch branch name is definitely not the thing I like to see. 😄

Hope we can keep pushing more zkVMs to satisfy the standard and finally stop dealing with this at the "guest program" level.

bin/stateless-validator-ethrex/sp1/Cargo.toml
```diff
@@ -7,26 +7,28 @@ version = "0.1.0"
...
+k256 = { git = "https://github.com/sp1-patches/elliptic-curves", tag = "patch-k256-13.4-sp1-6.0.0" }
+substrate-bn = { git = "https://github.com/sp1-patches/bn", tag = "patch-0.6.0-sp1-6.0.0-substrate-bn" }
 
-[patch."https://github.com/lambdaclass/bls12_381"]
-bls12_381 = { git = "https://github.com/lambdaclass/bls12_381-patch/", branch = "expose-fp-struct" }
+# [patch."https://github.com/lambdaclass/bls12_381"]
+# bls12_381 = { git = "https://github.com/lambdaclass/bls12_381-patch/", branch = "expose-affine-constructors" }
```
jsign: Do we need to create an issue to track this? Since this will prob have an impact in the EVM BLS precompiles perf. Worth pinging Ivan (Ethrex) about it?

bin/stateless-validator-reth/zisk/Cargo.toml
```diff
@@ -7,31 +7,27 @@ edition = "2024"
...
 # enable features
-alloy-primitives = { version = "1.5.0", default-features = false, features = [
+alloy-consensus = { version = "1.5", default-features = false, features = [
+    "crypto-backend",
+] }
+alloy-primitives = { version = "1.5", default-features = false, features = [
     "map-foldhash",
-    "sha3-keccak",
+    "native-keccak",
 ] }
```
jsign, line 16: I'm looking at this here: https://github.com/alloy-rs/core/blob/0fc5ab481952347a594984b47e7fa64a6a29016e/crates/primitives/src/utils/mod.rs#L184-L206

So basically zisk is always linking this `native_keccak256` now?

Feels a bit unfortunate that keccak (both for EVM and outside EVM-needs) is using `alloy-primitives` that ultimately isn't using this same "crypto backend" abstraction. I don't recall if there is a pending PR somewhere or is something to keep in mind suggesting/doing. cc @kevaundray

bin/stateless-validator-reth/openvm/src/openvm_revm_crypto.rs
```diff
@@ -1,43 +1,56 @@
-// Copied from https://github.com/axiom-crypto/openvm-reth-benchmark/blob/openvm-v1.4.1-reth-1.8.3/crates/revm-crypto/src/lib.rs
-
+//! Copied and modified from https://github.com/axiom-crypto/openvm-eth/blob/938d3c0/crates/revm-crypto/src/lib.rs.
```
jsign, line 1: I hope OpenVM can also start moving more into the C accelerators standard since this is hairy.

jsign (approved):
LGTM! Great PR, exciting to see more progress on crypto backends

crates/stateless-validator-ethrex/src/guest/zisk.rs
han0110: Created https://github.com/eth-act/ere-guests/issues/28

- For [Reth crypto backend](https://github.com/0xPolygonHermez/zisk-eth-client/tree/v0.7.0/crates/guest-reth/src/crypto), the reason we can't use directly is because they need some internal funcitons of `stateless` crate which is not exposed, after https://github.com/paradigmxyz/stateless/pull/12 is resolved we should be able to use it directly.
- For [Ethrex crypto backend](https://github.com/0xPolygonHermez/zisk-eth-client/tree/develop-0.8.0/crates/guest-ethrex/src/crypto), the reason is because they have bumped to `ziskos@pre-develop-0.17.0`, and there is some extern functions naming changes so we can't simply patch it back to `ziskos@v0.16.0`

I think it'd be nice if the crypto backend crate `zisk-{reth,ethrex}-crypto-backend` and both have releases `{reth_version,ethrex_version}-{zisk_version}` (just like the other patches), but I think it'd be hard to maintain all the combinations.

In current releases (both EL and zkVM runtime could be updated), if we are lucky, we can still depend on them and just patch the EL or zkVM runtime crate, but if there is incompatible API changes, we'd end up having to vendor them.

crates/integration-tests/src/lib.rs
han0110: > whatever the guest program wrote as public input isn't exactly the same thing we get from the SDK but padded

Yes, even we only write less than 256 bytes, the output is alwasy fixed 256 bytes (same for Airbender and OpenVM, but for them it's 32 bytes), but the unwritten parts will remain zeros. Agree that ideally  the output should be identical to what the guest writes.

bin/stateless-validator-ethrex/sp1/Cargo.toml
han0110: Dug deeper now I see that it's an optimization for the case of public key recovery (OpenVM also has the trick [here](https://github.com/eth-act/ere-guests/blob/han/feature/ere-v0.5.0/bin/stateless-validator-reth/openvm/src/openvm_revm_crypto.rs#L81) and [here](https://github.com/eth-act/ere-guests/blob/han/feature/ere-v0.5.0/bin/stateless-validator-reth/openvm/src/openvm_revm_crypto.rs#L290)), but for the p256 precompile we don't call the recovery function so it doesn't need to be patched, so I guess I'd just removed this line.

bin/stateless-validator-reth/zisk/Cargo.toml
han0110: I think for Reth all the `keccak256` routes to the one in `alloy-primitives`, and using the extern function is similar to crypto backend to me, just in this case we need to make sure the extern function name is correct, but no need to do patching (probably more error prone and symbol duplication could also happen).

For those who don't patch `sha3` or `tiny-keccak` (ZisK and OpenVM), they have the extern `native_keccak256` exported in their zkVM runtime.

I think in `libssz` we need something like this for `sha256` (or do another proper crypto backend abstraction to be installed explicitly).

bin/stateless-validator-reth/openvm/src/openvm_revm_crypto.rs
han0110: Can't use it directly because the `alloy-consensus` vesion the upstream dpends on is older which doesn't implement the `CryptoProvider::verify_and_compute_signer_unchecked` :(

bin/stateless-validator-ethrex/sp1/Cargo.toml
han0110: The reason it's commented out is because it's using old `sp1-lib`, which calls the extern function using `*{const,mut} u32`, but the v6 runtime already switch `*{const,mut} u64`, I'm not sure if it'd cause alignment problem, so I disabled it (although I tried to enable it and it can run 500 mainnet blocks + EEST 10M blocks, but still not confident that it's fine)

Add a note in https://github.com/eth-act/ere-guests/pull/27/commits/2ba20afc4c135f47188e56cf53680559be1b2d37

bin/stateless-validator-reth/zisk/Cargo.toml
jsign: > I think in libssz we need something like this for sha256 (or do another proper crypto backend abstraction to be installed explicitly).

Good point, I can ping Ivan later about this.

crates/integration-tests/src/lib.rs
marcinbugaj: Moving to standards should improve the situation. Zisk is actively working on adopting IO and accelerators standards.

PR #29, opened 2026-03-26
Update `ere` to `v0.6.0`

han0110:
- Update ZisK from `patch/v0.16.0` to `patch/v0.16.1`
- Update `libssz` to `v0.2.1` that provide interface to fill in sha256 impl, and implement the sha256 by extern function calls for OpenVM and ZisK
- Remove pico since it's removed in `ere` now


jsign (approved):
LGTM!

PR #30, opened 2026-03-27
Update to latest ethrex

jsign:
This PR updates Ethrex to the current latest commit, for two reasons:
1. Remove the fork we had that was waiting for https://github.com/lambdaclass/ethrex/pull/6392
2. Ethrex at that fork moment didn't use the Crypto provider to use keccak, thus all trie related work calling keccak wasn't using zkVM precompiles.

Did a quick `ziskemu` check and now the cost stops being blown up by the tries keccak load, which is how I noticed something was wrong reg point 2. above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2108,7 +2108,7 @@ dependencies = [
 [[package]]
 name = "ethrex-blockchain"
 version = "9.0.0"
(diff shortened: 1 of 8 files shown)


han0110 (approved):
LGTM

PR #31, opened 2026-04-03
Add stateless-validator-debug crate with CLI for running guest fixtures natively

jsign:
This PR:
- Add `stateless-validator-debug`, a CLI tool for running stateless validator guest fixtures directly on the host without needing a zkVM
- Supports both `--guest reth` and `--guest ethrex`, accepts a single JSON fixture file or a directory of fixtures
- Update README with the new crate

The main goal was to have a quick way to natively debug fixtures in guest programs. For example, I'm investigating some weird behaviour in a particular fixture, and it is quite helpful to have a way to run the guest natively, adding extra printlns or similar to figure out what is going on.

Usage:
```
cargo run -p stateless-validator-debug -- --guest reth path/to/fixtures/
cargo run -p stateless-validator-debug -- --guest ethrex fixture.json
```

### DIFF
--- a/crates/stateless-validator-debug/src/lib.rs
+++ b/crates/stateless-validator-debug/src/lib.rs
@@ -17,6 +17,7 @@ use stateless_validator_ethrex::guest::{
 use stateless_validator_reth::guest::{
     StatelessValidatorOutput, StatelessValidatorRethGuest, StatelessValidatorRethInput,
 };
+use tracing_subscriber::EnvFilter;
 
 /// CLI options for the stateless validator debug runner.
 #[derive(Debug, Clone, PartialEq, Eq, Parser)]
@@ -30,6 +31,9 @@ pub struct Cli {
     /// Guest program to run.
     #[arg(long, value_enum)]
     pub guest: GuestKind,
+    /// Warn and continue when fixture success does not match guest output.
+    #[arg(long)]
+    pub allow_success_mismatch: bool,
     /// Path to a fixture file or directory.
     pub path: PathBuf,
 }
@@ -138,6 +142,7 @@ impl Platform for StdoutNoopPlatform {
 
 /// Entry point for the debug runner binary.
 pub fn main_entry() -> anyhow::Result<()> {
+    init_tracing();
     execute(Cli::parse(), |summary| println!("{summary}"))
 }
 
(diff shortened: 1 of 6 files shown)


crates/stateless-validator-debug/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.expected_success,
+            self.actual_success,
+            encode_hex(&self.new_payload_request_root),
+        )
+    }
+}
+
+/// A no-op platform for host-side guest execution that forwards debug messages to stdout.
+#[derive(Debug)]
+pub struct StdoutNoopPlatform;
```
jsign, line 124: Doubted a bit about using `NoopPlaform`, which I created when I did the EIP-8025 PR. That would require adding the integration tests crate as a dependency, which is somewhat overkill. Also thought about moving `NoopPlatform` to an independent crate that is more neutral, but should also be `std`.

Ended up just creating a new one here since it is quite simple, and maybe even useful to be tunable independently of `NoopPlatform`.

No strong opinion, though.

crates/stateless-validator-debug/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        if summary.actual_success != summary.expected_success {
+            bail!(
+                "fixture {} ({}) expected success={}, got success={}",
+                summary.fixture_name,
+                fixture_path.display(),
+                summary.expected_success,
+                summary.actual_success,
+            );
+        }
```
han0110: Is this early return expected? Just thought it might be nice to run a single command to debug all fixtures in a dir and not stopped in the middle.

(we can always add a extra flag `--keep-going` later)

han0110 (approved):
LTGM! I had the exactly same tool locally and it's nice to have it here for reuse.

crates/stateless-validator-debug/src/lib.rs
jsign: Yep, I preferred an early return since it sounded like a safer behavior considering something happened that wasn't expected reg fixture expectations.

But I agree that allowing for continue can be useful. I added `--allow-success-mismatch` in https://github.com/eth-act/ere-guests/pull/31/commits/6b834f6c26749200ede81b531936a540a4dd8af6.

Issue #28, opened 2026-03-24
Use crypto backend from upstream for ZisK


han0110:
In #27 the crypto backend impls of ZisK are vendored, the reason:

- For [Reth crypto backend](https://github.com/0xPolygonHermez/zisk-eth-client/tree/v0.7.0/crates/guest-reth/src/crypto), the reason we can't use directly is because they need some internal funcitons of stateless crate which is not exposed, after https://github.com/paradigmxyz/stateless/pull/12 is resolved we should be able to use it directly.
- For [Ethrex crypto backend](https://github.com/0xPolygonHermez/zisk-eth-client/tree/develop-0.8.0/crates/guest-ethrex/src/crypto), the reason is because they have bumped to ziskos@pre-develop-0.17.0, and there is some extern functions naming changes so we can't simply patch it back to ziskos@v0.16.0

We should remove them and use the one from upstream directly once the issues are resolved.

han0110:
Closing this since ZisK uses generic crypto backends implemented by `zkvm_interface` now (#47)

PR #32, opened 2026-04-09
feat: remove ethrex zisk crypto

han0110:
- Update `ethrex` to rev `6ba9d9a458e812ff3afd80eaf89cddd97d3fb67e`
- Unvendor ZisK crypto. Resovles Ethrex part of #28


crates/stateless-validator-ethrex/Cargo.toml
```diff
@@ -51,8 +51,8 @@ default = ["std"]
 std = ["stateless-validator-common/std"]
 # NOTE: Need additional `ethrex-crypto/risc0` to make `KZG_PRECOMPUTE = 0`
 risc0 = ["ethrex-guest-program/risc0", "ethrex-guest-program/c-kzg", "ethrex-crypto/risc0"]
-sp1 = ["ethrex-guest-program/sp1"]
-zisk = []
+sp1 = ["ethrex-guest-program/sp1", "ethrex-guest-program/secp256k1"]
```
han0110, line 54: Enable `ethrex-guest-program/secp256k1` only to make the fallback impl inactive that's not compatible with SP1's `k256` patch, we still use `k256` in the crypto backend impl.

jsign (approved):

PR #35, opened 2026-04-14
BAL devnet 3 support

jsign:
This PR has changes to support `bal-devnet-3` branches for ELs.

High-level overview of changes:
- Switch Ethrex to `bal-devnet-3` plus some cherry-picked commits -- these commits are mainly required fixes that happened in `main` but haven't been merged in `bal-devnet-3` yet.
- Switch Reth stateless crate to `bal-devnet-3`, which also means Reth-crates too. It also has some extra patch I made, mostly reg allowing BAL as output in fixture runs, so we can recreate BALs for `ExecutionPayload` construction as we do with Requests. Seeing if I can upstream this soon to remove the patch.
- Add `NewPayloadRequestAmsterdam` to support Amsterdam accordingly (i.e. new `ExecutionPayload` types, etc). Also, correct deserialization and validations for both ELs, etc. 
- I changed the `blocks.tar.gz` for CI with some blocks I pulled from the real BAL devnet 3 chain that went live just some days ago.

Main goal here is:
- Have a branch to potentially prove bal devnet 3 to see BAL influence.
- Run EEST benchmarks filled for Amsterdam
- Have some things setups for potential sub-block proving experiments, which rely on BALs.

I think for a bit, this PR will be kept as a draft. At some point, we can think it makes sense to merge, but that can make things quite unstable, so I'm not totally sure about that.  We could use features to gate things, but it can get hairy -- I think we can discuss this eventually if it ends up being the main branch we use soon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,7 +286,7 @@ dependencies = [
  "rand 0.9.2",
  "rapidhash",
  "ruint",
- "rustc-hash 2.1.1",
+ "rustc-hash",
  "serde",
  "sha3",
 ]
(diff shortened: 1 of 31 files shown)


jsign:
Closing since I did https://github.com/eth-act/ere-guests/pull/36 with a branch owned by this repo which I think is better.

PR #36, opened 2026-04-14
Support BAL Devnet 3

jsign:
This PR has changes to support `bal-devnet-3` branches for ELs.

High-level overview of changes:
- Switch Ethrex to `bal-devnet-3`
- Switch Reth `stateless` crate to `bal-devnet-3`, which also means Reth-crates too. 
- Add `NewPayloadRequestAmsterdam` to support Amsterdam accordingly (i.e. new `ExecutionPayload` types, etc). Also, correct deserialization and validations for both ELs, etc. 
- I changed the `blocks.tar.gz` for CI with some blocks I pulled from the real BAL devnet 3 chain that went live just some days ago. Had to do a [bal-devnet-3 for workload](https://github.com/eth-act/zkevm-benchmark-workload/pull/276)

Main goal here is:
- Have a branch to potentially prove bal devnet 3 to see BAL influence.
- Run EEST benchmarks filled for Amsterdam
- Have some things setups for potential sub-block proving experiments, which rely on BALs.

I think for a bit, this PR will be kept as a draft. At some point, we can think it makes sense to merge, but that can make things quite unstable, so I'm not totally sure about that.  We could use features to gate things, but it can get hairy -- I think we can discuss this eventually if it ends up being the main branch we use soon.


.github/workflows/integration-test.yml
```diff
@@ -33,6 +33,10 @@ jobs:
             zkvm: airbender
           - guest: stateless-validator-ethrex
             zkvm: openvm
+          - guest: stateless-validator-reth
+            zkvm: risc0
+          - guest: block-encoding-length
+            zkvm: risc0
```
jsign, line 39: Looks like Risc0 can't be supported now reg the same point about Reth requirements. See previous [failing run](https://github.com/eth-act/ere-guests/actions/runs/24416075101/job/71325451273#step:9:949).

.github/workflows/compile-and-release.yml
```diff
@@ -55,6 +55,7 @@ jobs:
       - name: Compile guest
         run: |
           docker run \
+            -e OPENVM_RUST_TOOLCHAIN=nightly-2025-10-30 \
```
jsign, line 58: Reth `bal-devnet-3` started asking for 1.93.0.

jsign (commented):
@han0110, not ready for review or similar (although feel free to skim if you're curious).

Just leaving two particular comments here just fyi (nothing to do).

han0110:
> I think for a bit, this PR will be kept as a draft. At some point, we can think it makes sense to merge, but that can make things quite unstable, so I'm not totally sure about that. We could use features to gate things, but it can get hairy -- I think we can discuss this eventually if it ends up being the main branch we use soon.

Perhaps we can do the release like `execution-spec-tests` to release the ELF with BAL feature in a dedicated branch? So the main release can still be used for `zkboost`, and other branches can be used to do benchmark.

.github/workflows/integration-test.yml
han0110: Created https://github.com/eth-act/ere/pull/329 and it should resolve the issue (bumping risc0's rust version to 1.94.1)

PR #37, opened 2026-04-16
feat: bump `ere` to `v0.8.0`

han0110:
- Update `ere` to `v0.8.0` with serveral breaking changes:
  - `ere-platform-trait` is renamed to `ere-platform-core`, and the image tag extraction script is updated:
    ```
    cargo tree -p ere-platform-core \
      | head -n 1 \
      | sed -E -e 's/.*\?rev=(.{7}).*/\1/' -e 's/.*\?tag=v([0-9.]+).*/\1/'
    ```
    that extracts either revision or tag
  - `ere-io` is removed, and `Guest::Io` is replaced with `Guest::{Input,Output}: Encode + Decode` that uses `ere-codec` to implement a canonical encoding/decoding format for guest 
  - `ere-zkvm-interface` is renamed to `ere-prover-core` (also avoid confusion with the `zkvm-inferface.h` in `zkvm-standard`)input/output
- CI
  - Program is removed from the release, now `ere-server` takes raw ELF
  - Program verifying key is added in release as `{guest_name}.vk`, which can be used to initialized `zkVMVerifier`
  - Do `compile-and-release.yml` workflow dry-run (compile/keygen without publish)
  - Check duplicated crates on zkVM runtime

### DIFF
--- a/.github/workflows/compile-and-release.yml
+++ b/.github/workflows/compile-and-release.yml
@@ -85,7 +85,6 @@ jobs:
             --program-vk-path /output/${{ matrix.guest }}-${{ matrix.zkvm }}.vk
 
       - name: Upload artifact
-        if: github.event_name != 'pull_request'
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.guest }}-${{ matrix.zkvm }}
--- a/.github/workflows/compile-and-release.yml
+++ b/.github/workflows/compile-and-release.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - main
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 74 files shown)


.github/workflows/compile-and-release.yml
```diff
@@ -41,36 +46,52 @@ jobs:
...
             -v $PWD/output:/output \
             ghcr.io/eth-act/ere/ere-compiler-${{ matrix.zkvm }}:${{ steps.get_ere_tag.outputs.ere_tag }} \
             --compiler-kind rust-customized \
             --guest-dir /ere-guests/bin/${{ matrix.guest }}/${{ matrix.zkvm }} \
             --output-dir /output \
-            --elf-name ${{ matrix.guest }}-${{ matrix.zkvm }}.elf \
-            --program-name ${{ matrix.guest }}-${{ matrix.zkvm }}
+            --elf-name ${{ matrix.guest }}-${{ matrix.zkvm }}.elf
+
+      - name: Generate program VK
```
jsign, line 77: Nice!

.github/workflows/compile-and-release.yml
```diff
@@ -41,36 +46,52 @@ jobs:
...
+          docker run \
+            -e RUST_LOG=info \
+            -v $PWD/output:/output \
+            ghcr.io/eth-act/ere/ere-server-${{ matrix.zkvm }}:${{ steps.get_ere_tag.outputs.ere_tag }} \
+            --elf-path /output/${{ matrix.guest }}-${{ matrix.zkvm }}.elf \
+            keygen \
+            --program-vk-path /output/${{ matrix.guest }}-${{ matrix.zkvm }}.vk
 
       - name: Upload artifact
+        if: github.event_name != 'pull_request'
```
jsign: Is this intentional? Maybe still useful to upload to actions artifacts?

jsign (approved):
LGTM! just a nit question.

.github/workflows/compile-and-release.yml
han0110: Was afraid we might hit some resource limit, but we can try first and see. Enabled that in https://github.com/eth-act/ere-guests/pull/37/commits/b929595e75b465386e13a5b4e1e1bd3131a4978e

PR #38, opened 2026-04-21
Add spec compliant zilkworm guest program

developeruche:


jsign:
@developeruche, should this PR still be open reg what happened in:
- https://github.com/eth-act/zkevm-benchmark-workload/pull/291
- and what's going on in https://github.com/eth-act/zilkworm-stateless
?

cc @han0110  if also has any opinion on what we should do.

han0110:
It looks like https://github.com/eth-act/zilkworm-stateless impl also takes spec compatible input? If so I think we can republish the ELF from it for now, and switch to upstream until it supports spec compatible input?

Since zesu is also taking spec compatible input, so I'm feeling there is no reason to support customized input anymore (I'm also working on making `ethrex` and `reth` to take spec compatible input and will submit a PR soon).

jsign:
> so I'm feeling there is no reason to support customized input anymore

@han0110, highly agree -- let's try to push for this as much as we can. It will simplify a lot of infra and reasoning about things.

After we have everyone (or as many as possible) respecting the specs, we can truly focus on seeing if there is a better shape of spec inputs that is more optimal and walk in lockstep with implementations measuring peformance etc.

e.g. I'm prety sure the current `state` vector in the guest input being lexicographically ordered isn't the best thing possible. Prob makes sense to order in DFS order to optimize linear read while verifying the witness or something like that. I think would be a simple experiment to do to measure performance and compare with current spec baseline. But yeah, I think we can have more time to experiment with those things after we simplify a bit our lives in everyone syncing in accepting the same input, since we can't justify "this idea makes sense only for X team" and might need to impl the proposed optimization in 2 or 3 guest programs and come to a conclusion with avgs. 

Thinking out loud here.

han0110:
> I think we can have more time to experiment with those things after we simplify a bit our lives in everyone syncing in accepting the same input, since we can't justify "this idea makes sense only for X team" and might need to impl the proposed optimization in 2 or 3 guest programs and come to a conclusion with avgs.

Yeah makes sense.

developeruche:
> @developeruche, should this PR still be open reg what happened in:
> 
> * [feat: add Zilkworm execution client variant zkevm-benchmark-workload#291](https://github.com/eth-act/zkevm-benchmark-workload/pull/291)
> * and what's going on in https://github.com/eth-act/zilkworm-stateless
>   ?
> 
> cc @han0110 if also has any opinion on what we should do.

I think we can close this PR and use the release from https://github.com/eth-act/zilkworm-stateless instead

Currently, I have finished the implementation for Zisk and RISC0 and tested them; these guest program inputs are all spec-compliant, btw

https://github.com/eth-act/zilkworm-stateless might become not so useful should "zilkworm guest" become spec-compliant, but that would not happen anytime soon, as the zilkworm's development plan is not in this direction

jsign:
Cool, all this makes sense. Closing to keep things clean here!

PR #39, opened 2026-04-21
feat(ethrex): Use EIP-8025 feature to delegate as much logic into their guest program

jsign:
## Summary

This PR rewires `stateless-validator-ethrex` to use Ethrex's native EIP-8025 execution path instead of maintaining a local `NewPayloadRequest -> Block` conversion layer.

The main goal is to keep the guest thinner and align our input format with the upstream Ethrex interface, while reducing duplicate payload/parsing logic on our side.

## What changed

- updated Ethrex dependencies to the `jsign/ethrex` fork at `a485d97`, which includes the EIP-8025 guest entrypoint we need. I'll wait until https://github.com/lambdaclass/ethrex/pull/6516 gets merged so we can change to ethrex repo again before merging.
- replaced the old `StatelessValidatorEthrexInput` rkyv-based guest input with raw EIP-8025 wire bytes
- added a small `wire` module that encodes the input as `[ssz_len: u32 LE][ssz_bytes][rkyv_witness_bytes]` -- this isn't perfectly EIP-8025 compliant since their `ExecutionWitness` with EIP-8025 feature still uses rkyv. The next step for Ethrex is moving the `ExecutionWitness` to be spec compliant (thus use SSZ). But we're getting closer.
- changed the guest to call `execution_program_eip8025_bytes(...)` directly and return the upstream `new_payload_request_root` / `valid` result
- moved host-side input construction to a dedicated `build_eip8025_input(...)` helper
- derived SSZ encode/decode for the shared `NewPayloadRequest` types needed to serialize Electra/Fulu payloads
- removed the local `execution_payload.rs` and `new_payload_request.rs` conversion/validation code, which is now redundant with the Ethrex implementation
- updated debug and integration-test call sites to build the new wire input
- limited the Ethrex path to Electra/Fulu fixtures, which matches the current EIP-8025-supported flow -- this is mostly because the `NewRequestPayload` expected in Ethrex today only supports this fork.

## Why

This reduces maintenance in a sensitive part of the stack:

- less duplicated block/payload translation logic in this repo
- closer alignment with the upstream Ethrex guest interface
- clearer separation between wire encoding on the host side and validation/execution inside Ethrex
- smaller surface area for fork-specific payload handling bugs


crates/integration-tests/tests/stateless-validator-ethrex.rs
```diff
@@ -5,14 +5,24 @@ use guest::Guest;
...
     let fixtures = get_fixtures();
-    let inputs = fixtures.into_iter().map(|fixture| {
+    let inputs = fixtures.into_iter().filter_map(|fixture| {
+        let fork = determine_fork_name(
+            &fixture.stateless_input.chain_config,
+            fixture.stateless_input.block.header.timestamp,
+        );
+        if !matches!(fork, ForkName::Electra | ForkName::Fulu) {
+            return None;
+        }
```
jsign, line 24: Mainly because now we rely on whatever input is supported by Ethrex which is Electra/Fulu and prob forward from now on.

Should prob be okay until we do runs for older forks some day. 
Amsterdam support should be added soon for Ethrex to continue with forward forks.

crates/stateless-validator-ethrex/src/execution_payload.rs
jsign, line 1: Nice all this is gone.

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -1,161 +1,97 @@
...
 };
 
-/// Input for the Ethrex stateless validator guest program.
-#[derive(rkyv::Serialize, rkyv::Deserialize, rkyv::Archive)]
-pub struct StatelessValidatorEthrexInput {
-    /// New payload request data.
-    pub new_payload_request: NewPayloadRequest,
-    /// database containing all the data necessary to execute
-    pub execution_witness: ExecutionWitness,
-}
```
jsign, line 28: This is gone, since now the guest program receives raw bytes which are directly passed to Ethreum guest program entrypoint. See L21.

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -1,161 +1,97 @@
...
 }
 
-/// [`Io`] implementation of Ethrex stateless validator.
-pub type StatelessValidatorEthrexIo =
-    IoRkyv<StatelessValidatorEthrexInput, StatelessValidatorOutput>;
+impl core::error::Error for StatelessValidatorEthrexIoError {}
 
 /// [`Guest`] implementation for Ethrex stateless validator.
 #[derive(Debug, Clone)]
 pub struct StatelessValidatorEthrexGuest;
```
jsign, line 16: This got massively simplified, since now we rely on Ethrex logic doing all this stuff.

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -1,161 +1,97 @@
...
-                    P::print("Panic occurred during validation\n");
-                    StatelessValidatorOutput::new(new_payload_request_root, false)
-                }
-            }
-        }
-
-        #[cfg(not(feature = "std"))]
-        {
-            Self::compute_inner::<P>(input, new_payload_request_root, crypto)
-        }
```
jsign, line 118: Trying to assume no std, tracked by https://github.com/lambdaclass/ethrex/issues/6339

crates/stateless-validator-ethrex/src/new_payload_request.rs
jsign, line 1: Another nice deletion, also done by Ethrex guest program now.

crates/stateless-validator-ethrex/src/wire.rs
jsign, line 1: This is the encoding defined by Ethrex today which has an hybrid of using SSZ for the `NewPayloadRequest` and rkyv for `ExecutionWitness`.

Eventually this should be gone whenever it complies to the spec (which uses all SSZ).

Cargo.toml
```diff
@@ -85,12 +85,12 @@ stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "ed189a519
...
-ethrex-rpc = { git = "https://github.com/lambdaclass/ethrex.git", rev = "6ba9d9a458e812ff3afd80eaf89cddd97d3fb67e", default-features = false }
-ethrex-vm = { git = "https://github.com/lambdaclass/ethrex.git", rev = "6ba9d9a458e812ff3afd80eaf89cddd97d3fb67e", default-features = false }
+ethrex-common = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
+ethrex-crypto = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
+ethrex-guest-program = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
+ethrex-rlp = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
+ethrex-rpc = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
+ethrex-vm = { git = "https://github.com/jsign/ethrex.git", rev = "a485d9717e9ce5da6901361b9a04c8026d62fa68", default-features = false }
```
jsign: To be updated whenever https://github.com/lambdaclass/ethrex/pull/6516 gets merged.

jsign:
@han0110, despite is better to wait until Ethrex PR is merged, I think from this side should be okay to have your eyes for a review.

crates/stateless-validator-ethrex/src/guest.rs
han0110: Does it mean that the `deserialize_input` cost is not merged into `stf`?

han0110 (commented):
Nice simplifcation, looks much cleaner! Just got a small question

crates/stateless-validator-ethrex/src/guest.rs
jsign: The generic `deserialize_input` scope is still separate, but it only covers copying the raw input bytes now. The actual EIP-8025 decode/deserialization happens inside Ethrex’s `execution_program_eip8025_bytes`, so that part is now accounted under `stf`. 

I think this prob means we should rename `stf` for Ethrex now to something a bit more generic, since doesn't have the same meaning as `stf` before. I'll think some name and change it :)

jsign: [Changed to `run_validation`](https://github.com/eth-act/ere-guests/pull/39/commits/df8109dac56105bd068d8c09bfc736525c2de4cf) -- not sure if fully convinced, but seems to be a fairer name.

jsign:
@han0110, the https://github.com/lambdaclass/ethrex/pull/6516 pr was merged. Now switched `Cargo.toml` to point to ethrex main repo and branch. Reg code, nothing interesting has changed since your last comment (only this [nit change](https://github.com/eth-act/ere-guests/pull/39/changes/62ca670ff95e9fd77b1ded8651bdfb1c2aabb9b8) during the ethrex pr reviewing).

han0110 (approved):
LGTM

PR #42, opened 2026-04-23
Update Reth stateless to v2

jsign:
Update to the latest `paradigmxyz/stateless` Reth v2 with many implications:
- Update the Zisk custom crypto implementation for the new revm `Crypto` trait API, including `PrecompileHalt` and borrowed BLAKE2 inputs.
- As above, also related to OpenVM and SP1.
- Remove the `block-encoding-length` benchmark crate and guest binaries from the workspace, docs, CI, and integration tests.
- Pin the OpenVM Rust toolchain in compile/integration CI with `OPENVM_RUST_TOOLCHAIN=nightly-2025-10-30`.
- Had to regenerate a new `blocks.tar.gz` since Reth v2 changed the types for Block and Transactions (e.g. uses different encoding for values, for example, hex vs decimal), plus removed Compat support, thus the field in block changed from snake case to camel case). Empty block and invalid block are the same; the 10 mainnet blocks were changed to ones I pulled just now from mainnet using tentative https://github.com/eth-act/zkevm-benchmark-workload/pull/279)


.github/workflows/compile-and-release.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       fail-fast: false
       matrix:
         guest:
-          - block-encoding-length
```
jsign, line 29: Reth v2 removed the bincode-compat from their `Block` type (and other ones).
For `stateless-validator-reth` we are okay since EIP-8025 we pass our own `NewPayloadRequest` and not `Block` anymore.

But `block-encoding-length` was relying on it -- instead of working around this, I prefered to remove this guest program since this was originally done in early days of experimenting using SSZ as encoding, which ended up being our conclusion today anyway.

jsign:
@han0110, before (or after?) merging might worth doing a temp check in zkboost maybe with this new ere-guest branch and the workload one. This Reth V2 stuff came with a lot of surprises. Can wait for your green signal before we hit merge even if reviewed.

Edit: Also did a `stateless-validation-debug` run with a re-generated EEST with 8k benchmarks with the new workload PR, and all `expected_success` matched -- just to be more sure about things. Although I trust more the 10 mainnet blocks doing better coverage since those have more complexity than benchmark blocks reg execution witness, but all helps.

han0110:
Make sense, will create a zkboost PR on this branch and see if there is anything unexpected.

Opened https://github.com/eth-act/zkboost/pull/60, also tested locally and it looks fine.

han0110 (approved):
LGTM!

PR #43, opened 2026-04-27
Add partial Amsterdam logic

jsign:
This PR tries to push into `main` as much logic as possible from https://github.com/eth-act/ere-guests/pull/36, so we keep the diff with that PR as small as possible.

### DIFF
--- a/crates/stateless-validator-reth/src/host.rs
+++ b/crates/stateless-validator-reth/src/host.rs
@@ -4,7 +4,7 @@ use alloc::{format, vec::Vec};
 use std::sync::Arc;
 
 use alloy_consensus::Transaction;
-use alloy_eips::{Encodable2718, eip7685::Requests};
+use alloy_eips::{Encodable2718, eip7685::Requests, eip7928::BlockAccessList};
 use alloy_genesis::{ChainConfig, Genesis};
 use alloy_primitives::U256;
 use anyhow::Context;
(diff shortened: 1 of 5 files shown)


crates/stateless-validator-reth/src/host.rs
```diff
@@ -272,6 +278,60 @@ pub fn to_new_payload_request(
                 &requests,
             )
         }
+        ForkName::Amsterdam => {
+            let withdrawals = build_withdrawals()?;
+            // TODO(Amsterdam): use actual BAL that should be received from parameter.
+            let block_access_list = Tx::try_from(alloy_rlp::encode(BlockAccessList::default()))
```
jsign, line 284: The impl of this TODO would be the code happening in the BAL-devnet-3 branch/PR, since current EL dependencies doesn't have the logic to do this here now.

crates/stateless-validator-reth/src/new_payload_request.rs
```diff
@@ -78,6 +78,8 @@ where
         chain_spec.is_prague_active_at_timestamp(sealed_block.timestamp),
     )?;
 
+    // TODO(Amsterdam): Add Amsterdam-specific validation.
```
jsign, line 81: Ditto, since current EL branches don't have the code to check this.

han0110 (approved):
LGTM

PR #44, opened 2026-04-27
Tentative canonical guest program input support for Ethrex

jsign:
This PR adds tentative support for the canonical EEST `statelessInputBytes` payload as a guest program input to the Ethrex stateless validator, alongside the existing legacy `StatelessInput` flow.

The goal is to start exercising the Ethrex guest with the canonical format that EEST's `blockchain_test` fixtures emit, without breaking the existing legacy fixture path used by the rest of the stack.

Ideally in the future, we would like to use this new `Canonical` stateless input in `ere-guests` since basically simplifies a lot of gymnastics we do today reg input preparation. This `Canonical` style is literally receiving fully prepared SSZ inputs from EEST fixtures, and mostly passing them directly to the guest program.

It is only implemented for Ethrex since I created a PR to support it https://github.com/lambdaclass/ethrex/pull/6550.

Notes: 
- this PR shouldn't affect the current workload or zkboost, since we still support the now-named `Legacy` input style.
- Reth is out of scope for now -- they don't even support any kind of SSZ input yet, and trying to support `Canonical` in their current style would simply add more tech debt to our repo that should eventually be removed after.

### Host-side input construction (`stateless-validator-ethrex`)

- introduced an `Eip8025InputSource` enum with two variants:
  - `Legacy { stateless_input, valid_block }` — the existing path, unchanged in behaviour
  - `Canonical { ssz_input, chain_config }` — accepts EEST `statelessInputBytes` directly
- `build_eip8025_input(...)` now dispatches on the variant; the legacy path is moved to `build_legacy(...)` and a new `build_canonical(...)` rkyv-encodes the chain config and frames it together with the SSZ input
- extracted the `alloy_genesis::ChainConfig -> ethrex ChainConfig` conversion into a shared `to_ethrex_chain_config(...)` helper so both paths use the same logic
- added a new wire encoding `encode_eip8025_canonical(ssz_input, rkyv_chain_config_bytes)` producing `[ssz_len: u32 LE][ssz_bytes][cfg_len: u32 LE][rkyv_chain_config_bytes]`. This is a tentative format I created, but I think is quite solid. (`ChainConfig` isn't part of the specs, so this is why is passed "separately", but at least all the rest is proxied directly from the generated EEST fixture stateless input)

### Fixture loading (`stateless-validator-debug`)

- split fixture handling out into a new `fixtures.rs` module
- `StatelessValidatorFixture` now wraps a `FixtureInput::{Legacy, Canonical}` enum
- `load_fixtures(path)` auto-detects the JSON shape:
  - legacy `{ name, stateless_input, success }` → one fixture
  - EEST `blockchain_test` map → one canonical fixture per `(test, block)` with a populated `statelessInputBytes`
- `chain_config_for_test(...)` builds an `alloy_genesis::ChainConfig` from the EEST `network` + `chainid` + `blobSchedule` using `reth_chainspec::create_chain_config` and the `ef-tests` `ForkSpec`
- Amsterdam is constructed manually because the pinned Reth `stateless` revision doesn't expose it yet — temporary workaround until upstream catches up. This allows to try make progress in `ere-guests@main` instead of other branches.
- the Reth guest currently bails on canonical input (`reth guest does not yet accept EEST canonical SSZ input`); only Ethrex consumes it


crates/integration-tests/src/stateless_validator.rs
```diff
@@ -20,11 +20,15 @@ pub struct StatelessValidatorFixture {
 
 /// Returns the expected `StatelessValidatorOutput` for a given block hash in the fixtures, which
 /// was calculated by an independent implementation.
-pub fn get_stateless_validator_output(block_hash: B256, success: bool) -> StatelessValidatorOutput {
+pub fn get_stateless_validator_output(
+    block_hash: B256,
+    success: bool,
+    chain_id: u64,
```
jsign, line 26: Technically is needed for the spec [StatelessValidationResult](https://github.com/ethereum/execution-specs/blob/ec23140720d6a9257a907c470ba1874623bd7b50/src/ethereum/forks/amsterdam/stateless_ssz.py#L171).

This part of the API is a bit unstable, so not sure yet if this will be the final form. Opened a convo to explore this a bit more with [this doc](https://hackmd.io/@jsign/eth-zkevm-chainconfig-statelessinput).

For now sticking to the current specs -- we can easily adjust whenever that convo comes to some conclusion and we change the specs.

crates/integration-tests/src/stateless_validator.rs
```diff
@@ -33,47 +37,47 @@ fn expected_execution_payload_tree_roots() -> HashMap<B256, B256> {
     HashMap::from([
         (
             b256!("e4bd1c4dc22a58a0a9a8e789e2c54b4ace2d1ebc16a605c3976723b52fc011f1"),
-            b256!("45328434f812b65daa21b4e8a3d6440d0da95fbd95a6c10b0a28f081cab53bd5"),
```
jsign, line 36: These things had to be updated since `MAX_WITHDRAWALS_PER_PAYLOAD` had the wrong value compared to the specs. The specs define a non-consensus compliant value for some testing framework problems.

Until those are fixed, we should stick to what the spec choose to allow running EEST tests properly.

So, since `MAX_WITHDRAWALS_PER_PAYLOAD` changes, that affects `hash_tree_root` results, thus why these updates are here.

crates/stateless-validator-common/src/guest.rs
```diff
@@ -23,14 +29,21 @@ pub struct StatelessValidatorOutput {
     pub new_payload_request_root: [u8; 32],
     /// Stateless validation is successful or not.
     pub successful_block_validation: bool,
+    /// Chain ID from the stateless validation chain configuration.
+    pub chain_id: u64,
```
jsign, line 33: https://github.com/ethereum/execution-specs/blob/ec23140720d6a9257a907c470ba1874623bd7b50/src/ethereum/forks/amsterdam/stateless_ssz.py#L166-L171 (at least for now)

crates/stateless-validator-common/src/new_payload_request.rs
```diff
@@ -34,7 +34,7 @@ impl Sha256Hasher for NativeSha256Hasher {
 }
 
 /// Limits
-pub const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 16;
+pub const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 65536; // Re-defined in EIP-8025 at least temporarily.
```
jsign, line 37: https://github.com/ethereum/execution-specs/blob/ec23140720d6a9257a907c470ba1874623bd7b50/src/ethereum/forks/amsterdam/stateless_ssz.py#L40-L43

crates/stateless-validator-debug/src/fixtures.rs
jsign, line 1: TL;DR is that `stateless-validator-debug` now supports two fixtures:
- The usual workload fixture -- nothing changed (called `Legacy` now)
- Literal [fixtures from EEST releases](https://github.com/ethereum/execution-spec-tests/releases?q=zkevm&expanded=true)

For the later, basically each EEST test block has a `statelessInputBytes` and `statelessOutputBytes`. We use literally both to run guest and check correctness.

Ideally, asap, we can simply always use this canonical format which would be great. Simple format (just bytes), no manipulation in ere-guests, and all ELs receive the same. Meanwhile, we still support both.

crates/stateless-validator-debug/src/fixtures.rs
```diff
@@ -0,0 +1,330 @@
...
+    network: &str,
+    chain_id: u64,
+    blob_schedule: &BTreeMap<String, EestBlobParams>,
+) -> anyhow::Result<alloy_genesis::ChainConfig> {
+    // Construct Amsterdam chain config manually since Reth stateless crate
+    // doesn't have Amsterdam support in the main branch. To avoid being
+    // blocked by this, construct it manually. Eventually we can remove this.
+    if network == "Amsterdam" {
+        return Ok(amsterdam_chain_config(chain_id, blob_schedule));
+    }
```
jsign, line 266: Our guest program and execution witness specs are for Amsterdam. Since we use Reth a bit here to manipulate EEST fixture format parsing for ChainConfig stuff, we would technically need to use `stateless@bal-devnet-3`.

Prefered to stick to normal `stateless` branch, and manually construct for Amsterdam.

See #l1-zkevm-protocol thread I opened reg ChainConfig -- ideally if we can come to a conclusion, all this will be gone too since will be part of SSZ input.

crates/stateless-validator-debug/src/lib.rs
```diff
@@ -48,23 +53,41 @@ pub enum GuestKind {
...
-                let input = build_eip8025_input(&fixture.stateless_input, fixture.success)?;
+                let source = match &fixture.input {
+                    FixtureInput::Legacy(stateless_input) => Eip8025InputSource::Legacy {
+                        stateless_input,
+                        valid_block: fixture.success,
+                    },
+                    FixtureInput::Canonical(canonical) => Eip8025InputSource::Canonical {
+                        ssz_input: &canonical.ssz_bytes,
+                        chain_config: &canonical.chain_config,
+                    },
```
jsign, line 74: Depending if we have workload or EEST fixtures, we prepare the input accordingly.

crates/stateless-validator-debug/src/lib.rs
```diff
@@ -169,16 +187,25 @@ fn init_tracing() {
...
+    allow_fixture_mismatch: bool,
 ) -> anyhow::Result<()> {
+    if let Some(expected_output_bytes) = &summary.expected_output_bytes {
+        return handle_output_mismatch(
+            summary,
+            fixture_path,
+            expected_output_bytes,
+            allow_fixture_mismatch,
+        );
+    }
```
jsign, line 202: Reg checking the guest output -- if we have output bytes it means we ran an EEST fixture. Compare against the fixture `statelessOutputBytes` literally. If not, do what we did before.

crates/stateless-validator-ethrex/src/host.rs
```diff
@@ -18,11 +18,40 @@ pub use {
...
+        chain_config: &'a alloy_genesis::ChainConfig,
+    },
+}
+
 /// Builds the local EIP-8025 input buffer for the ethrex guest.
-pub fn build_eip8025_input(
-    stateless_input: &StatelessInput,
-    valid_block: bool,
-) -> anyhow::Result<Vec<u8>> {
+pub fn build_eip8025_input(source: Eip8025InputSource<'_>) -> anyhow::Result<Vec<u8>> {
```
jsign, line 41: Still single method to build Ethrex input, just that now we have these variants.

For workload/zkboost, they would just need to use `Legacy` as usual, and all should be okay. I think at some point we could make some helper to construct he canonical form  from the `eth_getBlock+debug_executionWitness` and we could switch over everything to `Canonical`.

But trying to go a bit in steps to avoid complicating things too agressively.

crates/stateless-validator-ethrex/src/host.rs
```diff
@@ -49,6 +78,75 @@ pub fn build_eip8025_input(
...
+    let cfg_bytes = rkyv::to_bytes::<rkyv::rancor::Error>(&ethrex_cfg)
+        .map_err(|err| anyhow::anyhow!("failed to rkyv-encode chain config: {err}"))?;
+
+    Ok(crate::wire::encode_eip8025_canonical(
+        ssz_input,
+        cfg_bytes.as_ref(),
+    ))
+}
+
+fn to_ethrex_chain_config(cfg: &alloy_genesis::ChainConfig) -> ChainConfig {
```
jsign, line 95: This is just extracted from below.

crates/stateless-validator-ethrex/src/wire.rs
jsign, line 1: Explained a bit in the PR description. Can also be useful to see https://github.com/lambdaclass/ethrex/pull/6550 but just "the expected thing" there.

Cargo.toml
```diff
@@ -86,16 +86,17 @@ reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", tag =
...
-ethrex-rpc = { git = "https://github.com/lambdaclass/ethrex.git", rev = "81484becd285e2a1e178306d8c22f94a104f560b", default-features = false }
-ethrex-vm = { git = "https://github.com/lambdaclass/ethrex.git", rev = "81484becd285e2a1e178306d8c22f94a104f560b", default-features = false }
+ethrex-common = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
+ethrex-crypto = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
+ethrex-guest-program = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
+ethrex-rlp = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
+ethrex-rpc = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
+ethrex-vm = { git = "https://github.com/jsign/ethrex.git", rev = "15ce41750a715177aef9f23a1adcd502b2e26166", default-features = false }
```
jsign: Mainly waiting to see what happens with https://github.com/lambdaclass/ethrex/pull/6550. When gets merged, can switch to latest Ethrex commit.

jsign (commented):
@han0110, I think this is ready for review without rush to gain a bit of time.

Still, for this PR to be merged we need to wait to see what happens with https://github.com/lambdaclass/ethrex/pull/6550. Maybe that happens next week. 

So mainly kind of half-opening for review. :)

crates/stateless-validator-common/src/guest.rs
```diff
@@ -86,7 +100,7 @@ impl Decode for StatelessValidatorOutput {
     type Error = StatelessValidatorOutputDecodeError;
 
     fn decode_from_slice(slice: &[u8]) -> Result<Self, Self::Error> {
-        if slice.len() < STATELESS_VALIDATOR_OUTPUT_SIZE {
+        if slice.len() != STATELESS_VALIDATOR_OUTPUT_SIZE {
```
han0110, line 103: The check was loose to support ZisK that has fixed size output 256 bytes, but since we always hash the output so this never gets called.

I think we can leave it stricter as is, and to be compatible with our zkVM standard IO interface in the future, I'm thinking we can probably force the output to be `[u8; 32]` so this `Decode` impl can be even removed.

crates/stateless-validator-debug/src/fixtures.rs
han0110: Nice!

han0110 (commented):
So far LGTM! Will check again when the Ethrex PR is merged

Cargo.toml
jsign: Got merged, so using lambdaclass repo main directly again.

crates/integration-tests/src/stateless_validator.rs
jsign: This got already resolved in the latest release -- instead of `chain_id : u64` we have a bit more comprehensive struct as input/output.

I think will take a bit of time for ELs to support this latest release, since it indirectly forces them to now support all changes in `bal-devnet-7`.

Leaving this as is now since it is the way that the latest v0.3.4 release works and most people are using/passing.

I soon can add a more focused PR supporting hte latest thing, or maybe if we are ambitious enough remove all this guest program wrappers and expect all ELs to return the output bytes themselves. Not sure yet how agressive we can be in shor term.

crates/stateless-validator-common/src/new_payload_request.rs
jsign: I also resolved this problem in latest zkevm@v0.4.0 -- for now leaving it compatible with v0.3.4 since it is the thing most people are using. As said before, I can update this soon to the original value, or maybe we can nuke everything here and expect this be a responsibility of the EL.

crates/stateless-validator-ethrex/src/host.rs
```diff
@@ -18,11 +18,40 @@ pub use {
...
+        stateless_input: &'a StatelessInput,
+        /// Whether the block is expected to validate successfully.
+        valid_block: bool,
+    },
+    /// EEST canonical `statelessInputBytes` SSZ payload + chain config.
+    Canonical {
+        /// SSZ-encoded EEST `statelessInputBytes`.
+        ssz_input: &'a [u8],
+        /// Chain config sourced from the fixture.
+        chain_config: &'a alloy_genesis::ChainConfig,
```
jsign, line 36: As mentioned before, in zkevm@v0.4.0, now `chain_config` will be part of `ssz_input`, so this will get simpler soon when more people start targeting that release 😄

jsign:
@han0110, added some boring extra commits compared to last review, plus some extra review comments (not that interesting either, just signaling good stuff happened pushing things in this repo to be simpler).

As we try to push ELs to stick to the latest standard, I think we can keep iterating on this changes topic to polish things better.

han0110 (approved):

PR #47, opened 2026-05-20
feat: update ere

han0110:
- Implement generic `ZkVMInterfaceCrypto` using `zkvm-interface` for both `reth` and `ethrex`, any [`zkvm-standard`](https://github.com/eth-act/zkvm-standards) compatible zkVMs can reuse
- Add extra ZisK guest ELF release with `cycle-scope` enabled with `-profiling` suffix to the file stem (e.g. `stateless-validator-ethrex-zisk-profiling.elf`), which can be used with `ziskemu -X -S -D ...` but can't be used to do proving


crates/downloader/src/lib.rs
```diff
@@ -261,12 +293,13 @@ mod tests {
 
     #[tokio::test]
     async fn download_from_tag() -> anyhow::Result<()> {
-        let guest = Downloader::from_tag("v0.9.0")
+        let guest = Downloader::from_tag("v0.11.0")
```
han0110, line 296: This is using the next release tag which will include the profiling ELF (currently CI is failing because it doesn't exist yet, but tested locally using the `Downloader::from_commit` it should be working.)

crates/stateless-validator-ethrex/src/guest/zkvm_interface.rs
```diff
@@ -0,0 +1,383 @@
+//! [`ethrex_crypto::Crypto`] implementation using [`zkvm_interface`].
```
jsign, line 1: So if any other zkVM-B implements the standard, we could in theory still use `zkvm_interface` from Zisk, as far as zkVM-B implements the C-funcs in some crate that is linked in compilation time, it should work, right? As far as I see `zkvm_interface` are only Rust-bindings, or has any other Zisk-specific detail?

I guess at some point we could have our own `zkvm_interface` crate in `eth-act`, so we can ask ELs to actually have all this code on their side, no? (No rush for this, just thinking out loud to check if makes sense).

bin/stateless-validator-reth/zisk/src/main.rs
```diff
@@ -2,28 +2,21 @@
...
-        deserialize_input,
-        new_payload_request_root_calculation,
-        misc_preparation,
-        new_payload_request_to_block,
-        stf,
-        serialize_output,
-        sha256_output_bytes,
-        write_output,
-    );
+#[unsafe(no_mangle)]
```
jsign, line 16: Is this required? Asking since I don't see this in the Ethrex main.

jsign (approved):
Nice! LGTM!

bin/stateless-validator-reth/zisk/src/main.rs
han0110: This is needed because we didn't turn on `once_cell/std` (for old Airbender to compile to standard target), and `reth` uses `once_cell` so we need to implement `critical_section`.

We can enable `once_cell/std` because Airbender also uses customized toolchain to compile and doesn't need `![no_std]`, but if we move to standard target with strict `![no_std]`, we will also need this again, so I just keep it here.

crates/stateless-validator-ethrex/src/guest/zkvm_interface.rs
han0110: > So if any other zkVM-B implements the standard, we could in theory still use `zkvm_interface` from Zisk, as far as zkVM-B implements the C-funcs in some crate that is linked in compilation time, it should work, right? As far as I see `zkvm_interface` are only Rust-bindings, or has any other Zisk-specific detail?


Yes, it should work in this way as long as the zkVM-B implements all the C-funcs. The actual impl of `zkvm_interface` of ZisK is in `ziskos` library that's only imported in the `bin` and get linked.

> I guess at some point we could have our own `zkvm_interface` crate in `eth-act`, so we can ask ELs to actually have all this code on their side, no? (No rush for this, just thinking out loud to check if makes sense).

Yes, I also think it'd be nice if we can host `zkvm_interface` somewhere with different lang bindings, so EL/zkVM share the same header. (perhaps just in the `zkvm-standard` repo so it stays synced)

PR #56, opened 2026-06-12
feat: cleanup guests

han0110:
- Remove `empty` and `panic` guests
- Rename `integration-tests` crate and move the test helper into `stateless-validator-test` crate
- Enable `ethrex-crypto/kzg-rs` in `stateless-validator-ethrex/host`


jsign:
Old guest programs: rest in peace :smile:

PR #58, opened 2026-06-16
feat: canonical stateless input

han0110:
This PR replaces the custom input serialization (`bincode` and `rkyv`) with the canonical SSZ-encoded `StatelessInput` following the `execution-specs@tests-zkevm@v0.4.1`, the idea is to have a Rust implementaton of `StatelessInput` we can reuse in `zkevm-benchmark-workload` and `zkboost`, and temporarily used to serve Ethrex and Reth guests until they are fully compatible.

## Changes

- Remove `trait Guest` abstraction to make guest libs simpler, now the guest libs simply export `fn entrypoint<P: Platform>()` that reads SSZed `StatelessInput` and writes SSZed `StatelessValidationResult`

### `stateless-validator-common`

- Implement `StatelessInput` and `StatelessValidationResult` base on `execution-specs@tests-zkevm@v0.4.1`, with backward support with older `NewPayloadRequest` variant (for `zkboost`)
- Remove custom input serialization derive (`serde` and `rkyv`)
- Support of legacy `reth_stateless::StatelessInput` is moved to `stateless_validator_common::host::legacy` (feature gated behind `legacy`).

### `stateless-validator-ethrex`

- Bump Ethrex to `b52aaea7c4b4e360a9d2e36aa624a4d1f4016011` to include https://github.com/lambdaclass/ethrex/pull/6863
- To support both current Mainnet (for `zkboost`) and Glamsterdam, we SSZ decode to our own `StatelessInput` and convert to `DecodedEip8025::Canonical` or `DecodedEip8025::Legacy`.
- Add host execution test

### `stateless-validator-reth`

- Bump Reth to `v2.3.0` and use [`han0110/stateless@feature/v2.3.0`](https://github.com/han0110/stateless/tree/feature/v2.3.0) patch with
  - In `zeth::SparseState` return error when missing witness node istead of panicking
  - Wire BAL check (hash and gas limit)
  - Skip sorting the ancestor headers and reject if not ordered
- SSZ decode to our own `StatelessInput` and convert to Reth input as is, and convert to Reth `ChainConfig` from canonical one.
- Add host execution test, but some fixtures are still failing, including 14 EIP-7610 and 1 EIP-8025 fixtures

### `stateless-validator-test`

- Move all zkVM execution tests into this crate since now they all take canonical input/output and no need custom input preparation.
- Add 3 preset of test fixtures
  - RPC fixtures of `bpo2` (current Mainnet)
    20 blocks (25234200-25234219) with `statelessOutputBytes` filled with `new_payload_request_root` calculated by `lighthouse`.
  - RPC fixtures of `glamsterdam-devnet-5`
    50 blocks (57050-57099) from [the R2 bucket of `glamsterdam-devnet-5` fixtures](https://pub-5345007fbd06486bbb7cbbe9f3112c45.r2.dev/devnets/glamsterdam-devnet-5/index.html) with `statelessOutputBytes` filled by `execution-specs@tests-zkevm@v0.4.1`
  - EEST fixtures from `execution-specs@tests-zkevm@v0.4.1` base on `bal-devnet-7`
- The RPC fixtures are temporarily hosted in https://github.com/han0110/ere-guests/releases/tag/rpc-fixtures@v0.1.0 to not blowup the git history, eventually we can remove them and only test with EEST fixtures I think
- zkVM execution test matrix
  - [`ethrex`] x [`risc0`, `sp1`, `zisk`] x [`rpc_bpo2`, `rpc_glamsterdam_devnet_5`, `eest_bal_devnet_7`]
  - [`reth`] x [`airbender`, `openvm`, `risc0`, `sp1`, `zisk`] x [`rpc_bpo2`, `rpc_glamsterdam_devnet_5`] (not yet passing all `eest_bal_devnet_7`)
  - [`zesu`] x [`zisk`] x [`eest_bal_devnet_7`] (`rpc_glamsterdam_devnet_5` is failing and pending to be resolved)

## Republish

- `zesu-zisk` - https://github.com/Consensys/zesu-zkvm/releases/download/bal-devnet-7-2026-06-12/stateless-validator-zesu-zisk.elf

## Future works

More things I'll do in a following PR after updating `ere`.

- Improve CI time
  - The integration testing time increases a lot for all except ZisK, this will be paritally fixed by updating the executor OpenVM and SP1 done in https://github.com/eth-act/ere/pull/383. And we can probably remove the non RV64 zkVM guests given that we already have 2 RV64 zkVMs (and the 3rd will arrive soon).
- Testing
  - Investigate why `zesu` doesn't pass `rpc_glamsterdam_devnet_5` fixtures and report if necessary.
- Republish
  - `zesu-zisk@bal-devnet-7-2026-06-24` that uses `zisk@v1.0.0-alpha`
  - `zesu-openvm@bal-devnet-7-2026-06-24` that depends on `ere` to update to `openvm@v2.1.0-dev.0` with RV64 support.
  - [`nethermind-zisk`](https://github.com/NethermindEth/nethermind/releases/tag/zisk-guest-r7) that uses `zisk@v1.0.0-alpha`


jsign:
@han0110, Haven't looked this PR in detail yet, but we have to remember if we always try to refer to the latest release (e.g today https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm%40v0.4.1) or `projects/zkevm` which might contain untagged changes. (Refresher: `projects/zkevm` contains actively working changes, and we cut releases in `projects/zkevm-releases` which usually is `projects/zkevm` rebased on top of some stable devnet branch (since `projects/zkevm` is based on `forks/amsterdam` which is a non-devnet "base" branch from `execution-specs`!).

This is relevant sometimes if there is some breaking change. For example, for the thing I commented yesterday of running glamstedam-devnet-5 artifacts in the spec input, I forgot to always try to target collection respecting (https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm%40v0.4.1) instead of `porjects/zkevm` and had a verification mismatch because of https://github.com/ethereum/execution-specs/pull/2926

That PR changes are not yet part of the latest release (i.e. NOT in v0.4.1), but I was using that `project/zkevm` for artifact collection. So when using v0.4.1 spec for validation had a failure of block validation because of that ChainConfig thing. (Edit: cross ref https://github.com/eth-act/zkevm-benchmark-workload/pull/294#discussion_r3424695946)

Again, I haven't dived into this PR yet so I'm not sure if this might be happening here. But I would say it is better to always try to target the latest release since usually most ELs are only targeting the latest release and not getting into the bleeding edge of `projects/zkevm` until those changes are tagged. Plus, I also think always targeting the latest release makes our life easier thinking of maintaining things.

Just leaving this here as food for though -- mostly because yesterday I spent 20m debugging why the spec couldn't validate the artifacts and was because this dummish mismatch.

Whenever a new tag is release, we might have to think how to reason about that in this repo since an immediate switch might make no guest satisfy any of the CI. So maybe we can have some policy of supporting last two releases or something. Honestly, I haven't thought about that but maybe worth we think about it just to have it clear and also make our lives easier and predictable.

All I said open for feedback!

han0110:
> But I would say it is better to always try to target the latest release since usually most ELs are only targeting the latest release and not getting into the bleeding edge of projects/zkevm until those changes are tagged. Plus, I also think always targeting the latest release makes our life easier thinking of maintaining things.

Agree on this, the reason this PR adds our own Rust impl of `StatelessInput` is also to be able to pin the spec version somehow (currently impl base on latest release `tests-zkevm/v0.4.1`), I found it's hard to debug specific guest when we found something diverge (e.g. new payload request root mismatch).

I think the cron testing in `zkevm-benchmark-workload` looks really nice, I think we can also try to add that so we have a confident Rust impl matching the `StatelessInput` pinned at specific version.

> That PR changes are not yet part of the latest release (i.e. NOT in v0.4.1), but I was using that project/zkevm for artifact collection. So when using v0.4.1 spec for validation had a failure of block validation because of that ChainConfig thing. (Edit: cross ref https://github.com/eth-act/zkevm-benchmark-workload/pull/294#discussion_r3424695946)

I used the `57050-57099` in this PR as the testdata and it has the spec fork number `24`, I think it's generated yesterday with the fix already?

---

Will polish the PR a bit and add details in the PR body soon, thanks for the note!

jsign:
> Agree on this, the reason this PR adds our own Rust impl of `StatelessInput` is also to be able to pin the spec version somehow (currently impl base on latest release `tests-zkevm/v0.4.1`)

Yes, I actually had to create the same structs in [workload here](https://github.com/eth-act/zkevm-benchmark-workload/blob/master/crates/witness-generator-spec-cli/src/schema.rs). Happy to nuke them whenever they get merged in this repo so we have a single place for Rust-like spec ones (maybe versioned by release tag). That would be clean :+1: 


> I found it's hard to debug specific guest when we found something diverge (e.g. new payload request root mismatch).

Yeah, I think what will happen is that we will have local ways of re-running the teams EL CI to build the ELF, and debug like that. Basically, locally building the ELF and inspecting why something isn't working. This is why them having transparent CIs that release ELFs will be very useful for us. Debugging just with an ELF is very hard.


> I think the cron testing in `zkevm-benchmark-workload` looks really nice, I think we can also try to add that so we have a confident Rust impl matching the `StatelessInput` pinned at specific version.

Yep. Now in workload is just running "plain Python spec" to verify. 
At some point we have to think where to do the same for the next step of using real ELFs via emulator and after if makes sense real proving with the cluster. I think that's the right order since via emulator also lives all in the realm of CI infra. The real proving maybe is more tricky, or doesn't make sense to do it with this cron archive history and give zkboost setup some airtime since that's was meant for RTP. e.g. zkboost glamsterdam devnet everyday from X to Y hours of day in cluster as fixed schedule. (Thinking out loud).

> > That PR changes are not yet part of the latest release (i.e. NOT in v0.4.1), but I was using that project/zkevm for artifact collection. So when using v0.4.1 spec for validation had a failure of block validation because of that ChainConfig thing. (Edit: cross ref [eth-act/zkevm-benchmark-workload#294 (comment)](https://github.com/eth-act/zkevm-benchmark-workload/pull/294#discussion_r3424695946))
> 
> I used the `57050-57099` in this PR as the testdata and it has the spec fork number `24`, I think it's generated yesterday with the fix already?

Yes this was exactly what I was looking at yesterday to check if you fall into the same trap I did or not, since I saw green CI. Luckly I looks like you used the artifacts I started generated after my fix, so all should be okay.

> Will polish the PR a bit and add details in the PR body soon, thanks for the note!

Nice, no rush! Stars are aligning...

Cargo.toml
```diff
@@ -64,66 +59,56 @@ tempfile = "3.6.0"
...
-stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "e5480e71a8031641ff471cad9dc482a7a14b32d5", default-features = false }
-tries = { git = "https://github.com/paradigmxyz/stateless", rev = "e5480e71a8031641ff471cad9dc482a7a14b32d5", default-features = false }
+reth-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
+reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
+reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
+reth-primitives-traits = { version = "0.4.1", default-features = false }
+reth-payload-validator = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
+reth-stateless = { git = "https://github.com/han0110/stateless", branch = "feature/v2.3.0", default-features = false, package = "stateless" }
+reth-tries = { git = "https://github.com/han0110/stateless", branch = "feature/v2.3.0", default-features = false, package = "tries" }
```
jsign, line 88: Is the diff too big? Plan maybe to create a PR?

Cargo.toml
```diff
@@ -64,66 +59,56 @@ tempfile = "3.6.0"
...
-libssz_types = { package = "libssz-types", version = "0.2.2", default-features = false, features = [
-    "alloc",
-] }
+libssz-derive = { version = "0.2.2" }
+libssz-merkle = { version = "0.2.2", default-features = false, features = ["alloc"] }
+libssz-types = { version = "0.2.2", default-features = false, features = ["alloc"] }
 
 # zkvm-standard
-zkvm-interface = { git = "https://github.com/0xPolygonHermez/zisk.git", tag = "v0.18.0" }
+zkvm-interface = { git = "https://github.com/eth-act/zkvm-standards", rev = "282cd356c3a0498416bb0619f9c8a347ce9933fb" }
```
jsign, line 102: Nice!

artifact-registry.json
```diff
@@ -1,5 +1,11 @@
 {
-    "stateless-validator-elf": {},
-    "stateless-validator-static-library": {},
-    "zkvm-static-library": {}
+    "stateless_validator_elf": {
```
jsign, line 2: This describe external "executable ELFs" that can be used?

artifact-registry.json
```diff
@@ -1,5 +1,11 @@
...
+        "zesu-zisk": {
+            "el_version": "bal-devnet-7-2026-06-12",
+            "url": "https://github.com/Consensys/zesu-zkvm/releases/download/bal-devnet-7-2026-06-12/stateless-validator-zesu-zisk.elf",
+            "sha256": "9d0db3d31f3f2342dce04dd87e31ae3efae40cb8ab03b911e581eae09cc7b4de"
+        }
+    },
+    "stateless_validator_static_library": {},
+    "zkvm_static_library": {}
```
jsign, line 10: 1. I imagine this `zkvm_static_library` will have some external link to a library per zkVM that we'll use to link relocatable ELFs to produce final ELFs?
2. What is `stateless_validator_static_library` field used for?

crates/stateless-validator-common/src/guest/input.rs
```diff
@@ -0,0 +1,373 @@
+//! Canonical stateless validation input types.
+//!
+//! The types mirror [`stateless.py`] and their SSZ schemas in [`stateless_ssz.py`]. The wire
+//! format is a 2-byte big-endian schema identifier followed by the SSZ-encoded `StatelessInput`
+//! container.
+//!
+//! [`stateless.py`]: https://github.com/ethereum/execution-specs/blob/tests-zkevm@v0.4.1/src/ethereum/forks/amsterdam/stateless.py
+//! [`stateless_ssz.py`]: https://github.com/ethereum/execution-specs/blob/tests-zkevm@v0.4.1/src/ethereum/forks/amsterdam/stateless_ssz.py
```
jsign, line 8: Cool, makes esnse. Worth having some subfolder with `v0.4.1` in the name now, or we can try to figure that out whenever I'll soon fire `v0.5.x` or whatever with `glamsterdam-devnet-5`?

Touching on this reg my comment of having at least two supported versions since prob we can't assume all guest program will instantly be ready for just-created releases.

I don't think we need to figure out this in this PR and can do it "as things happen" in the next `tests-zkevm@v...` release, so in a separate PR.

crates/stateless-validator-common/src/guest/input.rs
```diff
@@ -0,0 +1,373 @@
...
+///
+/// The spec fixes this schema to the Amsterdam payload shape. This
+/// implementation extends it and accepts payload shape from Bellatrix onward
+/// under the same identifier.
+pub const STATELESS_INPUT_SCHEMA_ID: u16 = 0x0001;
+/// Byte length of the big-endian schema identifier prefix.
+pub const STATELESS_INPUT_SCHEMA_ID_SIZE: usize = 2;
+
+/// SSZ list bounds from the Amsterdam stateless schema.
+pub const MAX_WITNESS_NODES: usize = 1 << 20;
```
jsign, line 33: Looks good -- some of these constants will change in the next `tests-zkevm@...` release, so will be nice to figure out how to organize. Just rumbling around the same comment as above. Nothing actionable now I think :)

crates/stateless-validator-common/src/guest/input.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub active_fork: ForkConfig,
+}
+
+impl ChainConfig {
+    /// Validates that the chain configuration is usable for the target payload, following
+    /// `validate_chain_config` in the spec with two deliberate differences. The spec's
+    /// blob-schedule equality check is performed by the verifier against the result public values
+    /// instead of here, and the spec's `fork != Amsterdam` rejection is generalized to a
+    /// payload-shape match through [`ProtocolFork::matches_payload`] for multi-fork inputs.
+    pub fn validate(&self, new_payload_request: &NewPayloadRequest) -> Result<(), Error> {
```
jsign, line 307: Makes sense. The `!= Amsterdam` from the specs is kind of related to `execution-specs` architecture, so the generalization is okay.

I think prob "soon" we'll nuke even this `validate(...)` stuff as we push more and more guest programs to provide to use a clear way to send the raw bytes directly and this live on their side.

But I understand today this is a good middleground to still adapt things!

crates/stateless-validator-common/src/guest/input/new_payload_request.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub execution_requests: ExecutionRequests,
+}
+
+/// Consensus-layer new payload request with one container shape per fork.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum NewPayloadRequest {
+    Bellatrix(NewPayloadRequestBellatrix),
+    Capella(NewPayloadRequestCapella),
+    Deneb(NewPayloadRequestDeneb),
+    ElectraFulu(NewPayloadRequestElectraFulu),
```
jsign, line 224: Once in a while think if we can nuke this since I'm still not sure most guest will support previous forks -- but seems like very low (or even zero) effort keeping this.

Will be useful if we can execute the idea of replaying the whole chain history generating guest inputs, and in that case we can use these to prepare things. So yeah, let's keep it although prob for some months might seem like dead code. Thinking out lout.

bin/stateless-validator-reth/airbender/src/main.rs
```diff
@@ -2,13 +2,13 @@
 
 #![no_main]
 
-use ere_platform_airbender::{AirbenderPlatform, entrypoint};
-use stateless_validator_reth::guest::{Guest, StatelessValidatorRethGuest};
+use ere_platform_airbender::AirbenderPlatform;
```
jsign, line 5: Don't remember last consensus about airbender chances of supporting the target we want -- but we maybe worth resuracing the discussion if this can simplify ere and ere-guests.

jsign: Thinking further: as teams guests provide relocatable ELFs and ere-guest main responsibility is just linking, maybe that will "auto prune" things.

bin/stateless-validator-reth/zisk/Cargo.toml
```diff
@@ -7,20 +7,20 @@ edition = "2024"
...
+alloy-consensus = { version = "2.0.5", default-features = false, features = [
     "crypto-backend",
 ] }
-alloy-primitives = { version = "1.5", default-features = false, features = [
+alloy-primitives = { version = "1.6.0", default-features = false, features = [
     "map-foldhash",
     "native-keccak",
 ] }
-revm = { version = "38.0.0", default-features = false, features = ["bn"] }
+revm = { version = "40.0.3", default-features = false }
```
jsign: Given this isn't activating `bn` anymore, could we remove this?

crates/stateless-validator-common/src/host/legacy.rs
```diff
@@ -0,0 +1,576 @@
...
+    },
+};
+
+impl StatelessInput {
+    /// Converts a legacy [`reth_stateless::StatelessInput`] into the canonical input.
+    ///
+    /// Configurations that merged through terminal total difficulty carry no
+    /// merge block, so Paris resolves as active since genesis and blocks
+    /// before the merge are not supported.
+    pub fn try_from_reth(stateless_input: &reth_stateless::StatelessInput) -> Result<Self> {
```
jsign, line 40: Uhm, so this is mostly to support the "old zkevm-benchmark-workload" fixture format only for Reth, no?

Maybe we could nuke it and just say we don't support Reth until they accept raw SSZ bytes?

Would that be "the tension" here? Because EEST spec fixtures already have the SSZ bytes, and for live networks we can already do spec-bytes with the zkevm-benchmark-workload style o using CL+EL RPCs.

Not saying we should change this PR -- mostly thinking if we are having this whole code only for Reth. Maybe the right incentive is to say Reth is gone until they support raw SSZ bytes as input and do the work on the `paradigmxyz/stateless` repo.

Am I missing something? LMK your thoughts

.github/workflows/integration-test.yml
```diff
@@ -79,29 +74,92 @@ jobs:
...
+      - name: Chown
+        if: matrix.external != true
+        run: |
+          sudo chown -R $(id -u):$(id -g) bin/stateless-validator-${{ matrix.guest }}/${{ matrix.zkvm }}/target || true
+
+      - name: Run sccache stat for check
+        shell: bash
+        run: ${SCCACHE_PATH} --show-stats
+
+  test_host_execution:
```
jsign, line 123: Nice, this can provide fast feedback

.github/workflows/compile-and-release.yml
```diff
@@ -113,6 +113,72 @@ jobs:
...
+          echo "ere_tag=${ERE_TAG}" >> $GITHUB_OUTPUT
+          echo "Ere tag: ${ERE_TAG}"
+
+      - name: Pull Ere images
+        run: |
+          docker pull ghcr.io/eth-act/ere/ere-server-${{ matrix.zkvm }}:${{ steps.get_ere_tag.outputs.ere_tag }}
+
+      - name: Download ELF
+        env:
+          ENTRY_KEY: ${{ matrix.el }}-${{ matrix.zkvm }}
```
jsign, line 146: Is the naming format used in Zesu releases feel fine or worth asking them to rename in a better shape?
Mostly asking since other ELs will jump into the same boat of releasing elfs so might be the right moment to suggest anything different (or at least just accept what it is in this PR, and share with them what we'll want if different)

.github/workflows/compile-and-release.yml
```diff
@@ -113,6 +113,72 @@ jobs:
...
+        run: |
+          ERE_TAG=$(cargo tree -p ere-platform-core | head -n 1 | sed -E -e 's/.*\?rev=(.{7}).*/\1/' -e 's/.*\?tag=v([0-9.]+).*/\1/')
+          echo "ere_tag=${ERE_TAG}" >> $GITHUB_OUTPUT
+          echo "Ere tag: ${ERE_TAG}"
+
+      - name: Pull Ere images
+        run: |
+          docker pull ghcr.io/eth-act/ere/ere-server-${{ matrix.zkvm }}:${{ steps.get_ere_tag.outputs.ere_tag }}
+
+      - name: Download ELF
```
jsign, line 144: OK, I see this connects to the `artifact-registry.json` new file -- makes sense.

So `artiact-registry.json` is mostly like a database for sources/origins, but the `republish` matrix definition is the intention, right?

Asking since the matrix and the .json file might not be 1:1, and just thinking if that makes sense.
Feels like a nit -- just sharing thought of the right way to "add a new guest": for now looks like editing both this `republish` matrix and the `artifact-registry.json` file, no?

.github/workflows/compile-and-release.yml
```diff
@@ -131,7 +197,7 @@ jobs:
 
   update-release-with-artifacts:
     name: Update release with artifacts
-    needs: [compile, check-git-tag]
+    needs: [compile, republish, check-git-tag]
```
jsign, line 200: Nice milestone!

crates/stateless-validator-test/src/execution/zkvm.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Returns path to the workspace root.
+fn workspace() -> PathBuf {
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    path.pop();
+    path.pop();
+    path
+}
+
+/// Resolves and caches guest ELF by compiling (ethrex, reth) or downloading (zesu).
+pub fn resolve_guest(guest_kind: GuestKind, zkvm_kind: zkVMKind) -> Elf {
```
jsign, line 26: Nice, makes sense. Hopefully we tends towards always `download_guest` 😄 Prob Ethrex can be the next move.

crates/stateless-validator-test/src/execution.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/// The public values written by a guest execution.
+#[derive(Debug, Clone)]
+pub enum ExecutionOutput {
+    /// The stateless output bytes.
+    Bytes(Vec<u8>),
+    /// The sha256 digest of the stateless output bytes.
+    Hash(Vec<u8>),
+}
```
jsign, line 49: I'll add a note in my task list to try to track this thing somehow as which zkVMs support raw bytes so we can simplify and remove `enum`. Prob will take time, but I think somewhere we should have some checklist of which zkVM supports vs doesn't support and up to which length etc and keep pushing to simplify. 

Just thinking out loud comment -- every time I see this raw vs sha256, I drop the same comment 😆

I think we'll start using Plane.so -- so we can start dumping all these little "remaining simplifications" in an ordered way so we can keep track of everything in a single place (soon).

crates/stateless-validator-test/src/execution.rs
```diff
@@ -0,0 +1,178 @@
...
+        match self {
+            Self::Bytes(stateless_output_bytes) => {
+                let expected_stateless_output =
+                    StatelessValidationResult::from_ssz_bytes(&expected_stateless_output_bytes)
+                        .map_err(|err| {
+                            anyhow::anyhow!("Decode fixture output bytes failure: {err:?}")
+                        })?;
+                if let Some((stateless_output_bytes, trailing)) =
+                    stateless_output_bytes.split_at_checked(expected_stateless_output_bytes.len())
+                    && trailing.iter().all(|byte| *byte == 0)
```
jsign, line 62: Yeah, and we should ask no trailing zero bytes requirement 😅

crates/stateless-validator-test/src/fixture.rs
```diff
@@ -0,0 +1,219 @@
...
+use tar::Archive;
+use tracing::info;
+use walkdir::{DirEntry, WalkDir};
+
+/// Release hosting the EEST fixtures filled by `ethereum/execution-specs`.
+const EEST_FIXTURES_BASE_URL: &str =
+    "https://github.com/ethereum/execution-specs/releases/download/tests-zkevm@v0.4.1";
+/// Release hosting the RPC-derived fixtures from `witness-generator-spec-cli`.
+const RPC_FIXTURES_BASE_URL: &str =
+    "https://github.com/han0110/ere-guests/releases/download/rpc-fixtures@v0.1.0";
```
jsign, line 22: Where these pulled fro mthe R2 bucket?

I imagine you did a mix of:
- Some pull from R2 bucket to get devnet-5 ones
- Some live RPC mainnet for BPO2 ones

Something like this?

We could try to automate this in a CI pipeline at some point in this repo maybe -- thinking out loud just to make this more transparent in how is generated, and once in a while refresh if needed (e.g. whenever `glamsterdam-devnet-6` is launched). (Not for this PR).

LMK your thoughts.

crates/stateless-validator-test/src/fixture.rs
```diff
@@ -0,0 +1,219 @@
...
+    dir: &'static str,
+    /// Subdirectory inside the unpacked archive holding the fixtures.
+    archive_dir: &'static str,
+}
+
+impl FixturePreset {
+    fn source(self) -> FixtureSource {
+        match self {
+            Self::EestBalDevnet7 => FixtureSource {
+                url: format!("{EEST_FIXTURES_BASE_URL}/fixtures_zkevm.tar.gz"),
```
jsign, line 49: Uhm nice, so we actually run all the EEST tests? Both for host and zkVM mode?
Would have imagined this was too slow.

So this is kind of what I did in https://eth-act.github.io/eest-execution-witness-dashboard/#/group/execution-witness but without Hive, no? (I mean for guest program part -- or executionWitness generation no since that's another story)

crates/stateless-validator-reth/tests/host_execution.rs
```diff
@@ -0,0 +1,18 @@
+//! Execution tests for `stateless-validator-reth` guest program on host
+
+use stateless_validator_reth::guest::run_stateless_guest;
```
jsign, line 3: Okay so `host` test run is on each guest crate in `host_execution.rs` and zkVM runs are in `stateless-validator-test` crate, no?

crates/stateless-validator-reth/src/guest.rs
```diff
@@ -1,148 +1,95 @@
...
-use stateless_validator_common::new_payload_request::{NewPayloadRequest, Sha256Hasher};
-use tries::zeth::SparseState;
-
-use crate::new_payload_request::new_payload_request_to_block;
-
-#[rustfmt::skip]
-pub use {
-    guest::*,
-    stateless_validator_common::{guest::StatelessValidatorOutput, new_payload_request},
+//! Reth stateless validator guest program.
```
jsign, line 1: Makes sense. Thinking out loud: ideal next move would be something like:
- Mostly "copy paste" this into `paradigmxyz/stateless`
- Make them have some CI to release some relocatable ELF

I don't think we should do it -- but we could write up some issue in their repo giving some directions or something like that so we can at some point nuke `stateless-validator-reth` stuff (and maybe more downstream things).

And maybe same issue text be an issue in Ethrex for them to do the same.

Does this sounds reasonable?

crates/stateless-validator-reth/tests/host_execution.rs
```diff
@@ -0,0 +1,18 @@
...
+// - 14 EIP-7610
+//    - test_init_collision_create_tx
+//    - test_init_collision_create_opcode
+//    - test_collision_with_create2_revert_in_initcode
+//    - test_create2_collision_storage
+// - 1 EIP-8037 (stale fixture, should be fixed in execution-specs#2892)
+//    - test_creation_tx_regular_check_subtracts_intrinsic_state
+// - 1 EIP-8025 (in-block created-code resolution)
+//    - test_witness_codes_create_same_hash_then_read
+declare_test_host_execution!(EestBalDevnet7, run_stateless_guest, should_panic);
```
jsign, line 18: Uhm, so this doesn't exactly check these specific fixtures are the ones failing no? Just lets all things fail for Reth for EEST tests?

crates/stateless-validator-ethrex/Cargo.toml
```diff
@@ -9,46 +9,26 @@ license.workspace = true
 workspace = true
```
jsign, line 9: No `host_execution.rs` in this `tests` crate as exists in Reth? Was this intentional?
Asking since maybe i'm confused

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -1,61 +1,99 @@
...
-    #[cfg(not(any(feature = "zkvm-interface", feature = "risc0", feature = "sp1")))]
-    return Arc::new(ethrex_guest_program::crypto::NativeCrypto);
+/// Validates the decoded payload through its canonical or legacy execution
+/// path, reporting a rejected payload as an error.
+fn run_validation(ethrex_input: DecodedEip8025) -> Result<(), ExecutionError> {
+    match ethrex_input {
+        DecodedEip8025::Legacy {
+            new_payload_request,
+            execution_witness,
+        } => validate_eip8025_execution(&new_payload_request, execution_witness, crypto::crypto()),
```
jsign, line 93: I think Ivan did some changes reg Legacy vs Canonical support.
Feels to me we should ask him a single entrypoint only accepting canonical and forget about legacy here.

I think this might be already be the case but i would have to check. If they support literall SSZ bytes from spec, feels like we can mostly nuke `stateless-validator-ethrex` stuf (e.g. convert.rs).

And if they publish elf in CI then the entire crate. But we can go in steps.

Thoughts? 😅

(Too many comments in this PR -- after is merged I can go through and create a single list o all TOODs or wishes and we can try to dump it in a single place and see next steps to ask external teams for stuff). Mostly at EOD the zesu guys did what we wanted from the start directly, no? (excluding relocatable ELF, but still...)

jsign (approved):
Excellent PR!

Left some comments to get your thoughts but I think nothing blocking. As said in some comment, after getting your replies and this being merged, I can go through all the comments again and dump TODOs/WISHES in a single doc so we can give clear asks/directions to guest teams to keep simplifying.

But this is a great step.

.github/workflows/compile-and-release.yml
han0110: > for now looks like editing both this republish matrix and the `artifact-registry.json` file, no?

Yes, to add a new republished guest we'd need to touch `artiact-registry.json` and the matrix here as well. The main reason is to speed things up, but it might make more sense to just read the json and iterate, so `artifact-registry.json` could be the single source of truth, the vk generation time seems to be negligible compare to compile, will update this in a following PR.

.github/workflows/compile-and-release.yml
han0110: I think the naming is fine to be anything in guest's release, since we'd only need a URL to download the ELF eventually. But it definitely makes it easier to identity which item to use from the release artifact list, if the naming is something like `stateless-validator-{el}-{zkvm}.elf` or `stateless-validator-{el}.a` or `stateless-validator-{el}.o`.

.github/workflows/integration-test.yml
han0110: Yes and I think it makes sense to do some more basic checks in `ere-guests` before publishing (so the Hive dashboard can have less surprise).

bin/stateless-validator-reth/airbender/src/main.rs
han0110: > as teams guests provide relocatable ELFs and ere-guest main responsibility is just linking, maybe that will "auto prune" things.

Yeah this will eventually make non-RV64 zkVM non-viable in `ere-guests`, but I'm thinking to removing them earlier considering we already have 2 RV64 zkVMs (and OpenVM RV64 might also land soon).

bin/stateless-validator-reth/zisk/Cargo.toml
han0110: Yes, nice catch!

crates/stateless-validator-common/src/guest/input/new_payload_request.rs
han0110: Was also not sure whether it's useful to support forks earlier than Mainnet, but I think it'd be good to at least keep current fork so we can run the Mainnet blocks, so we can still test zkboost on Mainnet spec (in case the EIP 8025 CLs are based on latest Glamsterdam spec).

crates/stateless-validator-common/src/guest/input.rs
han0110: Ah yeah that sounds good, will split into sub modules for a size 2 window for the latest specs

crates/stateless-validator-common/src/host/legacy.rs
han0110: > Because EEST spec fixtures already have the SSZ bytes, and for live networks we can already do spec-bytes with the zkevm-benchmark-workload style o using CL+EL RPCs.

Yes, the reason I still kept it here is just trying to be backward compatible in case the downstream usage still depends on Reth's stateless input, but as you said for `zkevm-benchmark-workload` it now has the spec bytes generator, and `zkboost` can adapt to the canonical input pretty easily, so I think we can nuke this soon.

crates/stateless-validator-ethrex/src/guest.rs
han0110: > I think this might be already be the case but i would have to check. If they support literall SSZ bytes from spec, feels like we can mostly nuke `stateless-validator-ethrex` stuf (e.g. convert.rs).

The legacy format is here mainly for Mainnet blocks support, because they only implement the canonical `StatelessInput` to Glamsterdam, but agree that i'd be better to have the Ethrex to support this directly. (I'm happy to upstream the changes but it might be a huge one).

> And if they publish elf in CI then the entire crate. But we can go in steps.

Yes and the last missing piece is to upstream the `Crypto` impl using `zkvm-standards`.

> Mostly at EOD the zesu guys did what we wanted from the start directly, no? (excluding relocatable ELF, but still...)

I think they did provide the relocatable ELF, but the problem to use them is they are using something outside of the `zkvm-standards`, I think we need to close this gap somehow (and probably in `ere-guests` we can try to have the shim crate to patch the gap temporarily like [this](https://github.com/han0110/reth-zisk-linking/tree/feature/sp1/shim))

crates/stateless-validator-ethrex/Cargo.toml
han0110: Yes there is [here](https://github.com/eth-act/ere-guests/pull/58/changes?mode=single#diff-27ea87a0e3cec6352fced641bb5936d2899588814b329c1c734a0b0645831512), and it passes all the fixtures we got currently

crates/stateless-validator-reth/src/guest.rs
han0110: Yes make sense, I'll create issues to report the gap and our current workaround to both Ethrex and Reth.

crates/stateless-validator-reth/tests/host_execution.rs
han0110: Yes, since eventually the `stateless-validator-{ethrex,reth}` will be nuked and we'll rely on the upstream to test spec compatibility I think.

crates/stateless-validator-reth/tests/host_execution.rs
han0110: These are the generators that generate the 16 failing fixtures (dig a bit and thought it'd be good to document here so I don't forget). The idea to still test it here (but mark it `should_panic`) is to share the failing fixtures easier like [this](https://github.com/eth-act/ere-guests/actions/runs/28148419529/job/83360444166?pr=58#step:7:392).

crates/stateless-validator-test/src/execution.rs
han0110: Yes I'm also pretty annoyed by this :(  It just adds an exception to handle on the harness (testing, benchmark, etc...).

I'm thinking if we can retain only the RV64 zkVM guests (SP1 and ZisK), we can actually skip the sha256, becuase ZisK support up to 256 bytes output, which is not `zkvm-standards` compatible but pretty enough for `StatelessValidationResult` (and if OpenVM RV64 lands it can also be configured to support fixed 256 bytes so we can still workaround it).

Do you think we should do this now?

crates/stateless-validator-test/src/execution.rs
han0110: Yeah, it's made mainly for ZisK's fixed size output (256 bytes)

crates/stateless-validator-test/src/fixture.rs
han0110: > I imagine you did a mix of:
> - Some pull from R2 bucket to get devnet-5 ones
> - Some live RPC mainnet for BPO2 ones

Yes exactly.

> We could try to automate this in a CI pipeline at some point in this repo maybe 

Yeah this sounds much better.

crates/stateless-validator-test/src/fixture.rs
han0110: Yes it's pretty slow for most zkVM (except ZisK which runs pretty fast).

In the next Ere release, the SP1 and OpenVM will also be improved a lot so it'd be doable to run the full EEST on Github runner within reasonable time.

> So this is kind of what I did in https://eth-act.github.io/eest-execution-witness-dashboard/#/group/execution-witness but without Hive, no? (I mean for guest program part -- or executionWitness generation no since that's another story)

Yes exactly, I'm doing this in `ere-guests` to get a better understanding of how compatible of the guests are, I feel it'd be safer at least test EEST and get a sense before releasing, so in Hive side we can have less surprises.

artifact-registry.json
han0110: Yes, this fields is for executable ELFs already (e.g. if guest teams don't release a relocatable ELF yet).

I think currently there is still some small blockers to "only do the linking", I'm most experimenting in [this repo](https://github.com/han0110/reth-zisk-linking) to figure out the gap.

artifact-registry.json
han0110: > 1. I imagine this `zkvm_static_library` will have some external link to a library per zkVM that we'll use to link relocatable ELFs to produce final ELFs?

Yes

> 2. What is `stateless_validator_static_library` field used for?

This is the field that stores the source of relocatable ELFs (e.g. [`zesu.rv64im.o`](https://github.com/Consensys/zesu/releases/download/bal-devnet-7-2026-06-24/zesu.rv64im.o)), and ideally we can do `zkvm_static_library` x `stateless_validator_static_library` linking to generate all the combinations.

han0110 (commented):
Thanks for the detailed review 🙏

Cargo.toml
han0110: Yes, will create a PR to `paradigmxyz/stateless` to update to `v2.3.0`

.github/workflows/compile-and-release.yml
jsign: Yep, single source of truth I think is less surprising but nbd or rush into it.

crates/stateless-validator-test/src/execution.rs
jsign: I would say to wait for OpenVM to catch up on allowing returning bytes to do this drop. I'm not sure yet how much we can rely on SP1 "counting" here. 

If Zisk and OpenVM can support raw bytes with "enough length", then as soon as that happens I think we can simplify here in `ere-guests`. And ofc try to ask them to support whatever max output was defined in the standards (which I don't remember from top of my head -- we me might need to revisit a bit things there to double check maximums and if they make sense)

crates/stateless-validator-test/src/fixture.rs
jsign: Agree with the strategy!

crates/stateless-validator-ethrex/src/guest.rs
jsign: Cool, sounds good. Maybe is just about us splitting all these things in some steps and ask or help them one by one in order of "simplifies our lives the most" to not overwhelm them with a huge thing (if possible). But Ethrex guys seems very open to do agressive changes.

crates/stateless-validator-test/src/execution.rs
han0110: Sounds good!

PR #59, opened 2026-06-30
feat: update ere

han0110:
- Dependencies
  - Bump `stateless` to `paradigmxyz/stateless`
  - Bump `ethrex` to `v19.0.0`
  - Bump `ere` to have `openvm@v2.0` with 256 bytes public values (configured by `ere`), and `sp1` with `zkvm-standards` impl
- Guests
  - `ethrex` and `reth` guests now output `stateless_output_bytes` instead of hash, since `openvm` and `zisk` they both support up to 256 bytes now (and `sp1` doesn't have limitation). And `zesu` and `nethermind` both output `stateless_output_bytes`, so the released guests are at least compatible on V4 payload fixtures (Amsterdam).
  - Drop `airbender` and `risc0` guests
  - Add `ethrex-openvm` guest
  - Use `zkvm-interface` crypto impl for `sp1` guests
  - Bump republished `zesu-zisk` to `zisk@v1.0.0-alpha` (only support V4 payload).
  - Add republished `nethermind-zisk` (support V3 payload by tweaking the schema id to `0x0000`)
- CI uses the `artifact-registry.json` as the single source of truth to do the `republish` (read the json and generate job matrix).

---

The blockers of linking guest lib and zkVM lib:

- `zisk` - The lib currently does allocation, but will be fixed in the next release (fixed by https://github.com/0xPolygonHermez/zisk/pull/1157)
- `sp1` - Same and will require some more changes in upstream to avoid the allocation.

So after `zisk@v1.0.0-beta` is released, we should be able to link `zesu` with `zisk` with a small shim (which relaxes the zkVM version matching a bit, and we only need to select the right EL version).


crates/stateless-validator-ethrex/src/guest/crypto/openvm.rs
```diff
@@ -0,0 +1,550 @@
+//! [`ethrex_crypto::Crypto`] implementation using OpenVM guest libraries.
```
jsign, line 1: We need this file since OpenVM does not provide some crate that will link the .so implementing `zkvm_interface`, no?

In any case, whenever we try to push Ethrex to release ELFs on their releases it is something that won't live in this repo anyway -- so prob our best strategy is simply trying to recommend them to publish ELFs and we simply republish, and move this on their side until OpenVM is ready.

I think this is already our plan -- repeating just to double check makes sense to you.

(BTW, supporting OpenVM for Ethrex now is great btw -- just thinking of next steps)

artifact-registry.json
```diff
@@ -1,9 +1,15 @@
 {
     "stateless_validator_elf": {
         "zesu-zisk": {
-            "el_version": "bal-devnet-7-2026-06-12",
-            "url": "https://github.com/Consensys/zesu-zkvm/releases/download/bal-devnet-7-2026-06-12/stateless-validator-zesu-zisk.elf",
-            "sha256": "9d0db3d31f3f2342dce04dd87e31ae3efae40cb8ab03b911e581eae09cc7b4de"
+            "el_version": "bal-devnet-7-2026-06-24",
+            "url": "https://github.com/Consensys/zesu-zkvm/releases/download/bal-devnet-7-2026-06-24/stateless-validator-zesu-zisk.elf",
```
jsign, line 5: Note: feels we should recommend the naem of the elf to contain the version of the zkVM.
A priori maybe not easy to see if this is for v1.0.0-alpha or similar.

Just maybe we could set some standard in naming to consider this. 
If this sounds like a good idea, I can add it to Plane as a task to discuss/define/publish eventually.

Cargo.toml
```diff
@@ -82,29 +76,41 @@ revm = { version = "40.0.3", default-features = false }
 reth-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
 reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
 reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-primitives-traits = { version = "0.4.1", default-features = false }
 reth-payload-validator = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-stateless = { git = "https://github.com/han0110/stateless", branch = "feature/v2.3.0", default-features = false, package = "stateless" }
-reth-tries = { git = "https://github.com/han0110/stateless", branch = "feature/v2.3.0", default-features = false, package = "tries" }
+reth-primitives-traits = { version = "0.4.1", default-features = false }
+reth-stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "2236845a84c3a7515312214e8a507e4731d73bb5", default-features = false, package = "stateless" }
```
jsign, line 81: Oh nice.

crates/stateless-validator-reth/src/guest/crypto/openvm.rs
```diff
@@ -1,20 +1,12 @@
+use openvm_sha2::{Digest, Sha256};
```
jsign, line 1: OpenVMV2 does not have sha256 precompile and we have to do naive?

crates/stateless-validator-reth/src/guest/crypto/sp1.rs
```diff
@@ -0,0 +1,8 @@
+#[unsafe(no_mangle)]
+extern "C" fn native_keccak256(bytes: *const u8, len: usize, output: *mut u8) {
```
jsign, line 2: Just to check if I understand correctly: this is just to adjust a naming mismatch between what extern keccak func Reth wants in their Crypto trait with the naming that we use in zkevm_interface (and SP1 sticks to). No?

crates/stateless-validator-test/src/execution/zkvm.rs
```diff
@@ -97,10 +128,25 @@ pub fn run_stateless_validator_execution(
...
-        })
+        let input = match guest_kind {
+            GuestKind::Nethermind => nethermind_input(input)?,
+            _ => input,
+        };
+        Ok(zkvm.execute(&Input::new().with_stdin(input))?.0.to_vec())
     })
 }
+
+/// Rewrites the schema prefix of `ElectraFulu` variant payload to `0x0000`.
```
jsign, line 139: Uhm, I see. Technically in the spec 0x0000 is not a valid value (was done intentionally to avoid mistakes i.e. avoiding default values just in case).

But I guess Nethermind assumed 0x0000 was for Fulu (since we started in the spec in Glamsterdam).

Is this correct?

I could meditate change 0x0002 for Glamsterdam and leave 0x0001 for Fulu so we can prove mainnet blocks.

Not a huge deal now. LMK your thoughts and if I'm understanding the situation correctly.

jsign (approved):
Nice! LGTM! Left some comments for your consideration, but nothing blocking.

crates/stateless-validator-ethrex/src/guest/crypto/openvm.rs
han0110: > We need this file since OpenVM does not provide some crate that will link the .so implementing `zkvm_interface`, no?

Yes.

> so prob our best strategy is simply trying to recommend them to publish ELFs and we simply republish, and move this on their side until OpenVM is ready.

Yes this makes sense to me, tho I think `openvm@v2.1` (RV64 support) will probably take some more time to land, that's why I still work on this to fill the guest gap.

Once `zisk@v1.0.0-beta` is released (the only one that's ready), I'll start start to drop the Rust guests in this repo. For `sp1` to be ready we'd need to update it to avoid heap allocation there. So hopefully we still have guests with more than 2 zkVMs.

crates/stateless-validator-reth/src/guest/crypto/openvm.rs
han0110: OpenVM doesn't expose a full `sha256` in `v2`, so we import that `openvm_sha2` which internally schedules the sha2 compressions (the compression is still by precompile)

crates/stateless-validator-reth/src/guest/crypto/sp1.rs
han0110: This is a leaking from the Reth that `zkvm-standards` didn't cover, in `alloy-primitives` if we don't want to patch the crate (`sha3` or `tiny-keccak`), we need to provide this `native_keccak256` so `alloy-primitives` routes the `keccak256` to it, and SP1's `zkvm-interface` impl doesn't contains that, so we need to patch it here (ZisK doesn't need one because `ziskos` exposes one)

If `stateless` repo eventually publishes a rellocatable ELF, this should be move to there so the ELF doesn't depend on symbol outside `zkvm-standards`

artifact-registry.json
han0110: The ELFs included here should be exactly same version used in the Ere we use here, that's why I didn't specify that.

But if the release name contains the zkVM version it'd be better to recognize for sure.

Tho eventually we don't need that since we'd link to specific zkVM version chosen by ourselves.

crates/stateless-validator-test/src/execution/zkvm.rs
han0110: > But I guess Nethermind assumed 0x0000 was for Fulu (since we started in the spec in Glamsterdam).

Yes

> I could meditate change 0x0002 for Glamsterdam and leave 0x0001 for Fulu so we can prove mainnet blocks.

I'm not sure in which case we'd bump the schema id, but if it's going to be bumped for each fork, leaving 0x0001 for Fulu and lifting Glamsterdam to 0x0002 sounds good to me (tho not sure if it's worth doing so since the Glamsterdam will land soon and Fulu support might not be necessary)

artifact-registry.json
jsign: Ah right, under a relocatable-ELF reality actually they ideal thing is they include the zkevm_standard version being supported

han0110: yeah that makes sense

crates/stateless-validator-test/src/execution/zkvm.rs
jsign: Uhm, yeah agree. Maybe better keep Glamsterdam = 0x0001 and keep schema-id as a structural signal of things rather than forks. At least that was my original intention, since for forks we already have `ChainConfig`.

So a bit two orthogonal things -- no need to get distracted by this 0x0000 ad-hoc value usage.

han0110: Yeah agree, even without the schema id hint, I think the guest can still decode the ExecutionPayload correctly by the fork passed in (I did that in https://github.com/eth-act/ere-guests/pull/59/changes/e0567c38a549edba373caa0957899a2d6a2b6172).

jsign: Yeah, maybe that's simpler for guest programs instead of the SSZ "magic" of reading some fields first as that commit.

So schema-id might end up being fork attached (plus potentially encoding attached, but doubt usage of this dimension will happen soon really; mostly a future proof thing that we won't be touching much reg encodings or extra stateless input fields for same fork)

PR #64, opened 2026-07-17
feat: update to tests-zkevm@0.6.2

han0110:
- Update `stateless-validator-common` to conform to `execution-specs@tests-zkevm@0.6.2`
- Bump `ethrex` rev to `e8860d2918921603afa67f62687e9a39f197367e`
- Bump `reth` to branch [`glamsterdam-devnet-7`](https://github.com/paradigmxyz/reth/tree/glamsterdam-devnet-7) (forked to lower the MSRV, might need for a while until SP1 and ZisK Rust fork is updated to >1.95)
- Bump RPC fixtures to adapt to new stateless input schema format
- `zesu-zisk` stay at `execution-specs@tests-zkevm@0.4.1` since no new release yet
- Add sanity check of `new_payload_request_root` between `stateless-validator-common` and `ethrex` when `debug_assertions` is enabled (enabled in CI testing).


crates/stateless-validator-test/src/execution.rs
```diff
@@ -107,47 +113,52 @@ pub fn run_execution(
...
-            })
-    else {
-        bail!(
-            "Output bytes mismatch, expected {}, got {}",
-            const_hex::encode_prefixed(expectecd_bytes),
-            const_hex::encode_prefixed(got_bytes)
-        )
+macro_rules! declare_matches_output {
+    ($name:ident, $stateless_validator_common_crate:ident) => {
+        fn $name(got_bytes: Vec<u8>, expectecd_bytes: Vec<u8>) -> anyhow::Result<()> {
```
jsign: `s/expectecd/expected`.

crates/stateless-validator-ethrex/src/guest.rs
```diff
@@ -54,18 +54,30 @@ pub fn run_stateless_guest<P: Platform>(input_bytes: &[u8]) -> Vec<u8> {
...
-fn verify_stateless_new_payload<P: Platform>(input: StatelessInput) -> Result<(), Error> {
+fn verify_stateless_new_payload<P: Platform>(
+    fork: ProtocolFork,
+    input: StatelessInput,
+) -> Result<(), Error> {
     P::cycle_scope("validate_chain_config", || {
         input.chain_config.validate(&input.new_payload_request)
     })?;
 
+    #[cfg(debug_assertions)]
```
jsign, line 65: Ah I see -- after we merge this PR, I'll ask Ivan to them working in releasing the ELF and we can forget about maintaining these "conversion checks". I think given devnet-7 is the last devnet and things won't change much, is a good moment for Ivan and Figtrace to make that move.

crates/stateless-validator-test/src/execution.rs
```diff
@@ -107,47 +113,52 @@ pub fn run_execution(
...
-            "Output chain_config mismatch, expected {:?}, got {:?}",
-            expected.chain_config,
-            got.chain_config
-        ),
-        _ => bail!("Output mismatch, expected {expected:?}, got {got:?}"),
-    }
 }
+
+declare_matches_output!(matches_output, stateless_validator_common);
+declare_matches_output!(matches_output_legacy, stateless_validator_common_legacy);
```
jsign: Uhm, I see. So we need this since Zesu isn't v0.6.2 ready yet.

Maybe at some point we can think if this makes sense for us to maintain, or whenever we start using a new devnet in ere-guests, then only include guest programs that are fully ready for it.

For externally compiled ELFs (which should be all hopefully soon), should be quite simple to remove or add them.

Maybe in the future we can make workload and ere-guests to be somewhat aligned in releases, and only support latest devnet.

If some client is behind in the previous devnet, then they will exist in the previous release.

So at workload and ere-guests we only care about ELFs for single devnets.

Thinking out loud here a bit. We can see with time -- it will definitely help this being the last devnet for G.

crates/stateless-validator-test/src/fixture.rs
```diff
@@ -45,21 +45,21 @@ struct FixtureSource {
...
         match self {
-            Self::EestBalDevnet7 => FixtureSource {
-                url: format!("{EEST_FIXTURES_BASE_URL}/fixtures_zkevm.tar.gz"),
-                dir: "eest-bal-devnet-7",
+            Self::EestGlamsterdamDevnet7 => FixtureSource {
+                url: format!("{EEST_FIXTURES_BASE_URL}/tests-zkevm@v0.6.2/fixtures_zkevm.tar.gz"),
+                dir: "eest-glamsterdam-devnet-7",
+                archive_dir: "fixtures/blockchain_tests",
+            },
+            Self::EestGlamsterdamDevnet5 => FixtureSource {
```
jsign: Right makes sense reg Zesu being one devnet behind.

Cargo.toml
```diff
@@ -63,28 +63,29 @@ walkdir = "2.5.0"
...
 
 # reth
-reth-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-payload-validator = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
-reth-primitives-traits = { version = "0.4.1", default-features = false }
-reth-stateless = { git = "https://github.com/paradigmxyz/stateless", rev = "2236845a84c3a7515312214e8a507e4731d73bb5", default-features = false, package = "stateless" }
-reth-tries = { git = "https://github.com/paradigmxyz/stateless", rev = "2236845a84c3a7515312214e8a507e4731d73bb5", default-features = false, package = "tries" }
+# FIXME: Switch back to a released tag once glamsterdam-devnet-7 lands upstream and zkVM compiler verison is bumped to 1.95.
```
jsign, line 76: Yeah. Or maybe we hsould ask Fig to do the release ELF pipeline and just include Reth when fully ready to avoid having this mainteinance burden.

crates/stateless-validator-test/src/execution.rs
```diff
@@ -107,47 +113,52 @@ pub fn run_execution(
...
+            let Some(got_bytes) = got_bytes.split_at_checked(expectecd_bytes.len()).and_then(
+                |(got_bytes, trailing)| trailing.iter().all(|byte| *byte == 0).then_some(got_bytes),
+            ) else {
+                bail!(
+                    "Output bytes mismatch, expected {}, got {}",
+                    const_hex::encode_prefixed(expectecd_bytes),
+                    const_hex::encode_prefixed(got_bytes)
+                )
+            };
+
```
jsign: Ah I see, this field by field comparison is mainly for better debugging instead of some opaque byte disagreement.
Very good idea!

jsign (approved):
LGTM! 🙏

crates/stateless-validator-ethrex/src/guest.rs
han0110: Yeah it'd be good if Etherx can take a look, I've been thinking how to mention the gap in an issue, was wondering if they can just check the code directly. The gap for Etherx should be small, mainly just the canonical chain config conversion and crypto provider via `zkvm-standards`. And since they already have the [guest bins](https://github.com/lambdaclass/ethrex/tree/main/crates/guest-program/bin) setup, it'd be feasible for them the release the ELF now I think.

For Reth I think it might take longer tho, especially Reth bumps MSRV to 1.95 and SP1 and ZisK still support up to 1.94, it might need to creative way to do the compilation (for ZisK we could probably compile via stock compiler + custom json target, for SP1 we probably need to use their pre-released 1.96 or use compile-and-link approach but rely on extra `sys_alloc_aligned`).

crates/stateless-validator-test/src/execution.rs
han0110: yeah I found it hard to compare raw bytes by my own eyes (tho raw bytes work fine for agents).

crates/stateless-validator-test/src/execution.rs
han0110: yeah I think that'd be the most clean approach, I'll just comment out zesu in this RP then. I kept it mainly because I remember we had a discussion to have keep 2 devnet supports, but probably not much useful since the tool depends on ere-guests will be confused (e.g. not knowing which guests are up-to-date).

Removed in https://github.com/eth-act/ere-guests/pull/64/commits/6ca84f88a8b8281f6bbcb3bfb94784fcbe8bd91f

Cargo.toml
han0110: yep that'd be nice, doing it here just don't want to be blocked.

crates/stateless-validator-test/src/execution.rs
han0110: Fixed in https://github.com/eth-act/ere-guests/pull/64/commits/198fb14e9038ca06533d3ce20262bdc875a9467b

PR #67, opened 2026-07-23
feat: add zesu and R2 glasmterdam-devnet-7 fixtures

han0110:
- Add `zesu-zisk` back of the [`glamsterdam-devnet-7.2`](https://github.com/Consensys/zesu-zkvm/releases/tag/glamsterdam-devnet-7.2) release
- Bump `ere` so ZisK gets the rigth Rust toolchain to compile guests
- Add selected 10 blocks from [glamsterdam-devnet-7 R2 bucket](https://pub-df22334654034ebab51bc096137a59d8.r2.dev/devnets/glamsterdam-devnet-7/index.html) as preset `RpcGlamsterdamDevnet7` and test
- Add cron workflow to test latest 100 blocks from the same R2 bucket every day.

### DIFF
--- a/.github/workflows/integration-test-devnet.yml
+++ b/.github/workflows/integration-test-devnet.yml
@@ -71,7 +71,6 @@ jobs:
         run: echo "ERE_PROFILE=$ERE_PROFILE" >> "$GITHUB_ENV"
 
       - name: Run zkVM execution of the latest devnet blocks
-        shell: bash
         env:
           RUST_LOG: info,ere_server=error
           ERE_IMAGE_REGISTRY: ghcr.io/eth-act/ere
@@ -79,7 +78,8 @@ jobs:
         run: |
           cargo run --release --package stateless-validator-test --bin zkvm_execution_devnet -- \
             --zkvm ${{ matrix.zkvm }} \
-            --stateless-validator ${{ matrix.stateless_validator }} | tee failures.txt
+            --stateless-validator ${{ matrix.stateless_validator }} \
+            --output failures.txt
           {
             echo "### ${{ matrix.stateless_validator }} on ${{ matrix.zkvm }}"
             echo '```'
--- a/.github/workflows/integration-test-devnet.yml
+++ b/.github/workflows/integration-test-devnet.yml
@@ -1,0 +1,97 @@
+name: Integration test devnet
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test_zkvm_execution_devnet:
+    name: ${{ matrix.stateless_validator }} on ${{ matrix.zkvm }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        stateless_validator:
+          - ethrex
+          - reth
+        zkvm:
+          - openvm
+          - sp1
+          - zisk
+        include:
+          - { zkvm: zisk, external: true, stateless_validator: zesu }
+          - { zkvm: zisk, stateless_validator: ethrex, ere_profile: ethrex }
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Free up disk space
+        if: matrix.zkvm == 'zisk'
+        run: bash .github/scripts/free-up-disk-space.sh
+
+      - name: Install Rust toolchain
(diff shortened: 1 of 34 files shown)


.github/workflows/integration-test-devnet.yml
```diff
@@ -0,0 +1,97 @@
...
+          path: bin/stateless-validator-${{ matrix.stateless_validator }}/${{ matrix.zkvm }}/target
+          key: ${{ format('{0}-bin-{1}-{2}-{3}', runner.os, matrix.stateless_validator, matrix.zkvm, hashFiles(format('bin/stateless-validator-{0}/{1}/Cargo.lock', matrix.stateless_validator, matrix.zkvm))) }}
+          restore-keys: |
+            ${{ runner.os }}-bin-${{ matrix.stateless_validator }}-${{ matrix.zkvm }}-
+
+      - name: Enable Ere profile
+        if: matrix.ere_profile != ''
+        env:
+          ERE_PROFILE: ${{ matrix.ere_profile }}
+        run: echo "ERE_PROFILE=$ERE_PROFILE" >> "$GITHUB_ENV"
```
jsign, line 71: Reminder for future: these `ere_profile` stuff will prob live in the Ethrex (or others) repo using whatever compilation flags per zkVM makes sense. So prob our CIs can become more agnostic about all this configuration surface.

.github/workflows/integration-test-devnet.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Run zkVM execution of the latest devnet blocks
```
jsign, line 73: Nice, this is great. Similarish to the scheduled checker of the R2 in workload but instead of using EEST, use the concrete guests and zkVMs.

.github/workflows/integration-test-devnet.yml
```diff
@@ -0,0 +1,97 @@
...
+          OPENVM_RUST_TOOLCHAIN: nightly-2026-01-18
+        run: |
+          cargo run --release --package stateless-validator-test --bin zkvm_execution_devnet -- \
+            --zkvm ${{ matrix.zkvm }} \
+            --stateless-validator ${{ matrix.stateless_validator }} \
+            --output failures.txt
+          {
+            echo "### ${{ matrix.stateless_validator }} on ${{ matrix.zkvm }}"
+            echo '```'
+            cat failures.txt
```
jsign, line 86: Not for this PR/now: At some point we can automate some dashboards to see all scheduled runs and show these failures in a single dashboard.
Not a big rush to do this, but maybe is a good idea as a daily check and ping ELs about failures or similar.

crates/stateless-validator-ethrex/tests/host_execution.rs
```diff
@@ -4,4 +4,5 @@ use stateless_validator_ethrex::guest::run_stateless_guest;
 use stateless_validator_test::declare_test_host_execution;
```
jsign, line 5: host-mode runs prob will go away whenever Ethrex moves to publishing ELFs, but still nice to have this mode until then anyway.

We can think of proposing them (in their repo!) to have host-mode runs instead of here (which won't be possible).

Something we can figure out eventually, since host mode runs are very nice for easy debug.

jsign (approved):
LGTM!

.github/workflows/integration-test-devnet.yml
han0110: Ah yeah a dashboard like Hive will be better to navigate the errors

crates/stateless-validator-ethrex/tests/host_execution.rs
han0110: yeah that makes sense, and actually Ethrex has host execution already now (e.g. the [runner](https://github.com/lambdaclass/ethrex/blob/main/tooling/ef_tests/blockchain/test_runner.rs#L604) and [action](https://github.com/lambdaclass/ethrex/actions/runs/30055530160/job/89366294213#step:6:15288)), but they only verify the `success` byte but not `new_payload_request_root` (probably it's not checked in the past due to no ssz, but now they should be able to check it).

.github/workflows/integration-test-devnet.yml
han0110: Yeah, just triggered a run and we can see the failures like [this](https://github.com/eth-act/ere-guests/actions/runs/30062059730#summary-89385507657), will add host execution there as well so we know whether it's a guest issue or a zkVM issue (also remove the sccache stats noise).

PR #75, opened 2026-08-06
feat: use openvm zkvm accelerator from ere

han0110:
- Use `zkvm_accelerator.h` impl exported from `ere-platform-openvm`
- Cleanup OpenVM specific crypto backend impls and use the shared `ZkVMInterfaceCrypto`


bin/stateless-validator-ethrex/openvm/src/main.rs
```diff
@@ -3,8 +3,6 @@
 use ere_platform_openvm::OpenVMPlatform;
 use stateless_validator_ethrex::guest::entrypoint;
 
-openvm::init!();
```
han0110, line 6: This is now enforced in https://github.com/eth-act/ere/blob/master/crates/platform/openvm/src/zkvm_accelerator.rs#L66 since the declaration depends on the `openvm.toml` config, but in ere we hardcode the stanard config.

bin/stateless-validator-reth/openvm/src/openvm_revm_crypto.rs
```diff
@@ -1,600 +0,0 @@
-//! OpenVM Crypto Implementation for REVM
```
han0110, line 1: Removed and use the shared `ZkVMInterfaceCrypto`, same for the ethrex one.

